PR #113, opened 2016-01-26
Fix some autobuild related issues

stweil:


egorpugin:
You are removed autobuild/autotools itself. Is it intended?
So, do you propose to use only cmake?

egorpugin:
Or those files actually always generated by autotools?
Sorry in this case. I misunderstood you.

stweil:
All those files are built by ./autobuild.

DanBloomberg:
Forgive me if I misunderstand -- I'm just starting to learn about git, and
Egor is helping me during the transition.

The master at
     https://github.com/DanBloomberg/leptonica/
has autobuild (unchanged from 1.72)

It also looks like it has all the autotools generated files: configure,
aclocal.m4, files in config/, and src/Makefile.in and prog/Makefile.in.

I downloaded the release and built it using autotools.  The only problem
was that configure wasn't executable; needs to be 755.

  -- Dan

DanBloomberg:
Yes, I did misunderstand.

Stefan, I see that you added those filenames to .gitignore.  But those
files are specific to the current set of source files and must be generated
and packed into the release.  We don't typically expect users to generate
them, just to ./configure (etc).

The situation seems different with cmake, where the analogous configuration
files are generated by the user when running cmake.

  -- Dan

stweil:
Yes, source code distributions should include those files because not everybody has the autotools installed.

For git repositories, the situation is different. Normally free software projects which use git don't include generated files in the repository. They assume that users who use the git repository also know how to use the autotools.

When you want to build a source distribution (tar.gz), all you have to do is to run ./autobuild first.

chewi:
As the resident autotools guy, I have to agree with @stweil. Rather than rely on GitHub's implicit releases through tagging, you'll have to prepare the tarball manually and upload it via the **Draft a New Release** button. I haven't done this before but it looks straightforward.

stweil:
@egorpugin, autobuild was not removed. I only added the missing execute permissions to that file.

egorpugin:
It's up to Dan.
As I understand you should run 'autobuild' command in order to create makefiles, isn't it?

stweil:
Yes, that's the usual procedure for git based builds.

DanBloomberg:
That's fine with me -- when in Rome ...    Later I'll add another shell
script that can also be used to run autotools.

PR #116, opened 2016-01-26
Partially fix JPEG 2000 support for Windows

stweil:
Now reading JP2 files works without admin rights on Windows, so it
is possible to build Tesseract for Windows with JPEG 2000 support.

### DIFF
--- a/src/jp2kio.c
+++ b/src/jp2kio.c
@@ -785,6 +785,19 @@ PIX      *pix;
 #if HAVE_FMEMOPEN
     if ((fp = fmemopen((void *)data, size, "rb")) == NULL)
         return (PIX *)ERROR_PTR("stream not opened", procName, NULL);
+#elif defined(_WIN32)
+    static const char fn_template[] = "mkstemp.XXXXXX";
+    static char filename[64];
+    char *tmpdir = getenv("TMP");
+    sprintf(filename, "%s/%s", tmpdir ? tmpdir : ".", fn_template);
+    int handle = mkstemp(filename);
+    L_WARNING("work-around: writing to a temp file\n", procName);
+    if ((fp = fdopen(handle, "r+b")) == NULL) {
+        printf("mkstemp stream not opened, %s\n", strerror(errno));
+        return (PIX *)ERROR_PTR("mkstemp stream not opened", procName, NULL);
+    }
+    fwrite(data, 1, size, fp);
+    rewind(fp);
 #else
     L_WARNING("work-around: writing to a temp file\n", procName);
     if ((fp = tmpfile()) == NULL)


DanBloomberg:
Stefan, does the call using tmpfile() not work for Windows?  I thought it
did.  e.g.,
    http://stackoverflow.com/questions/6247148/tmpfile-on-windows-7-x64
or is the problem that it still makes it in the root directory:
    http://www.elpauer.org/2011/08/the-most-stupid-c-bug-ever/

Of all the functions for getting a temporary file, this is the only one (at
least in Unix) that is not deprecated.  That is because it just returns a
file stream, not a file name, so it's relatively safe from exploits.

  -- Dan

stweil:
tmpfile() for Windows tries to create a temporary file in the root directory (usually c:\). This used to work long ago, but nowadays the root directory needs special permissions, so file creation will fail there for normal users (administrators won't have a problem). See also https://msdn.microsoft.com/de-de/library/x8x7sakw.aspx.

DanBloomberg:
Stefan, Egor,

We've got 5 instances of this in the library.  I added a function to
utils.c to handle fmemopen and the work-arounds, and used it wherever we
were reading from memory.

I don't have my git set up, so I've attached a tar with 7 files.  Please
put it into the master.

The complementary function for writing to memory via a file stream is
unfortunately not so easily separated from its use.  (I assume that's what
you meant by 'partially fix'.)

One of the files in the tar is allheaders.h, which has all the function
prototypes (and in this case, a new one for fopenReadFromMemory).
This is generated from my static makefiles with 'make allheaders', which
runs:
      xtractprotos -protos=inline -prestring=LEPT_DLL *.c
It would be nice if cmake produced a similar target  (e.g., 'make
allheaders'), because you do not want to generate modifications to
allheaders.h with an editor.

Thanks.

  -- Dan

stweil:
That's right – not only JPEG 2000 uses that workaround with a temporary file, so either my modification has to be used several times, or you can use a new utility function. I added it for JPEG 2000 because that was my primary focus then.

If Leptonica would support image files in memory (or memory mapped files) instead of the stream based API, we could get rid of the temporary files and get better performance, too. All relevant host platforms support memory mapped files. Is there a chance to go into that direction without breaking backward compatibility?

DanBloomberg:
On Wed, Jan 27, 2016 at 1:42 PM, Stefan Weil <notifications@github.com>
wrote:

> That's right – not only JPEG 2000 uses that workaround with a temporary
> file, so either my modification has to be used several times, or you can
> use a new utility function. I added it for JPEG 2000 because that was my
> primary focus then.
>

​Made sense to write the read-from-memory utility function.​

​ I also need to do something on the write-to-memory side.​

> If Leptonica would support image files in memory (or memory mapped files)
> instead of the stream based API, we could get rid of the temporary files
> and get better performance, too. All relevant host platforms support memory
> mapped files. Is there a chance to go into that direction without breaking
> backward compatibility?
>

​Haven't thought about it.  File systems are pretty fast.  Maybe we can
defer​

​on this ...

DanBloomberg:
New version of tar file attached for fixing the tmpfile problem on windows,
both reading from and writing to memory.  Made an additional function in
utils.c just for windows (so now there are two new functions there).

Needless to say, I haven't even compiled the windows versions of this
code.  Used "w+b" when opening for write, in analogy to "r+b" that you
used.  However, *nota bene*:  Stefan, I see that in your original patch,
you opened a temp file to write with "r+b".  I copied this in the attached
code.  The documentation for flags to fdopen says:

*r* or *rb*Open a file for reading.*w* or *wb*Open a file for writing.*a*
 or *ab*Open a file for writing at end-of-file.*r*+ or *rb*+ or *r*+*b*Open
a file for update (reading and writing).*w*+ or *wb*+ or *w*+*b*Open a file
for update (reading and writing).*a*+ or *ab*+ or *a*+*b*Open a file for
update (reading and writing) at end-of-file.

which seems a bit strange (r+b and w+b appear to be the same)!

In any event, if this works, the issue should be solved.

  -- Dan

stweil:
@DanBloomberg, `r+b` opens an existing file for read/write, `w+b` opens an existing file for read/write and destroys any previous contents. Here `r+b` is sufficient because the file was just created and still is empty.

egorpugin:
This PR was commited with changes to master. ca9d16af5

DanBloomberg:
Egor,

No, please roll that back.  It violates C89.  The tar I sent you does this
properly, for ALL the image io files, and it also handles the writing to
memory part.

  -- Dan

egorpugin:
Roll what back?
I did not commit THIS pull request.
I commited YOUR changes from tar.gz.

DanBloomberg:
Egor, that's fine.  I'm still learning git.  When someone makes a pull
request and you do a "merge", that's not a commit?

To give you an idea of the depth of my present confusion, here's some
documentation on git merge, from
    http://www.sbf5.com/~cduan/technical/git/git-3.shtml

*Let the current head be called current, and the head to be merged called
merge. *

   1. *Identify the common ancestor of current and merge. Call it
   ancestor-commit.*
   2. *Deal with the easy cases. If the ancestor-commit equals merge, then
   do nothing. If ancestor-commit equals current, then do a fast forward
   merge.*
   3. *Otherwise, determine the changes between the ancestor-commit and
   merge.*
   4. *Attempt to merge those changes into the files in current.*
   5. *If there were no conflicts, create a new commit, with two parents,
   current and merge. Set current (and HEAD) to point to this new commit, and
   update the working files for the project accordingly.*
   6. *If there was a conflict, insert appropriate conflict markers and
   inform the user. No commit is created.*

Yup, clear as mud.

  -- Dan

stweil:
Dan, in this case it was quite easy: Egor took the updated source files from your tar file and committed them with git. Then he closed my pull request. "This PR was commited with changes to master" did not mean that there was a git merge. It just means that you copied some of my code / ideas in your code, and that was committed.

DanBloomberg:
Stefan, thanks.  In the next couple of days I'll have my local source set
up for git and will do as you suggest with respect to the extern
declarations.

Did you see the appveyer failure in win32 and win64, where it apparently
couldn't find mkstemp()?  (There was also an error due to my mis-spelling
the new function fopenWriteWinTmpfile!).

*C:\projects\leptonica\src\utils.c(1876): error C3861: 'mkstemp':
identifier not found [C:\projects\leptonica\build\src\leptonica.vcxproj]*
*C:\projects\leptonica\src\utils.c(1921): error C3861: 'mkstemp':
identifier not found [C:\projects\leptonica\build\src\leptonica.vcxproj]*

Any idea why we got that error?

  -- Dan

egorpugin:
There's no such function in Microsoft's C Runtime (in Visual Studio).
"To create a temporary file in a directory other than the root, use tmpnam or tempnam in conjunction with fopen."
https://msdn.microsoft.com/en-us/library/x8x7sakw.aspx

By the way, windows builds are tested using Visual Studio 2015 and not mingw or cygwin.
Mingw/cygwin will work just fine if linux builds work find (travis-ci tests that).

egorpugin:
And just to be clear leptonica cannot be compiled with C89 compiler at the moment.
It can be compiled with C89 plus GNU extensions.
I tried following options: ansi,c89,c90,c99,c11 - nothing compiles.
But it compiles with non standard GNU extensions in GCC: gnu89,gnu90,gnu99,gnu11 - all compiles.

So, you should keep it in mind when talking about C89 compatibility. Or you should fix the code and we'll add compiler option, so checking will be strict.

stweil:
I'll prepare a patch to fix compilation with Visual Studio.

PR #120, opened 2016-01-30
Remove unneeded extern declarations

stweil:

### DIFF
--- a/src/bmpio.c
+++ b/src/bmpio.c
@@ -569,9 +569,6 @@ RGBA_QUAD  *pquad;
 /*---------------------------------------------------------------------*
  *                         Read/write to memory                        *
  *---------------------------------------------------------------------*/
-#if HAVE_FMEMOPEN
-extern FILE *open_memstream(char **data, size_t *size);
-#endif  /* HAVE_FMEMOPEN */
 
 /*!
  *  pixReadMemBmp()
--- a/src/jp2kio.c
+++ b/src/jp2kio.c
@@ -743,9 +743,6 @@ opj_image_cmptparm_t  cmptparm[4];
 /*---------------------------------------------------------------------*
  *                         Read/write to memory                        *
  *---------------------------------------------------------------------*/
-#if HAVE_FMEMOPEN
-extern FILE *open_memstream(char **data, size_t *size);
-#endif  /* HAVE_FMEMOPEN */
 
 /*!
  *  pixReadMemJp2k()
--- a/src/jpegio.c
+++ b/src/jpegio.c
@@ -940,9 +940,6 @@ jmp_buf                      jmpbuf;  /* must be local to the function */
 /*---------------------------------------------------------------------*
  *                         Read/write to memory                        *
  *---------------------------------------------------------------------*/
-#if HAVE_FMEMOPEN
-extern FILE *open_memstream(char **data, size_t *size);
-#endif  /* HAVE_FMEMOPEN */
 
 /*!
  *  pixReadMemJpeg()
--- a/src/pngio.c
+++ b/src/pngio.c
@@ -1254,9 +1254,6 @@ l_pngSetReadStrip16To8(l_int32  flag)
 /*---------------------------------------------------------------------*
  *                         Read/write to memory                        *
  *---------------------------------------------------------------------*/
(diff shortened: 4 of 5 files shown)


DanBloomberg:
No, don't do this.  We need the open_memstream functions in all of those
files when compiling with gcc.

Let me repeat: *with respect to 1.73*,
*just use the tar file I sent yesterday for all changes to:*

*bmpio.c*

*jpegio.c*

*jp2kio.c*

*pngio.c*

*pnmio.c*

*utils.c*
*allheaders.h*

That declares fmemopen() in utils.c and open_memstream() in the 5 above
*io.c files.

Do not use Stefan's changes to those files.  They will not compile with gcc
and C89.

  -- Dan

stweil:
The patch was already applied to git master.

Dan, if you use library functions which are defined in system header files, then those files have to be used. Never write your own `extern` declaration. Doing that is not the right way to fix compilation problems. When do you see failing compilations without those declarations? For me on Debian GNU Linux both gcc and clang work fine with my changes.

DanBloomberg:
Running on ubuntu 14-4 with gcc:
      CC = gcc -ansi -D_BSD_SOURCE -DANSI -fPIC
without the extern declaration I get this (in bmpio.c):

bmpio.c:587:13: warning: assignment makes pointer from integer without a
cast [enabled by default]
     if ((fp = fmemopen((l_uint8 *)cdata, size, "rb")) == NULL)
             ^

The compiler is apparently expecting an int to be returned, instead of a
FILE*.

The prototype in stdio.h is guarded by a variable __USE_XOPEN2KB, whatever
that is.

stweil:
That's caused by the `-ansi` option. Just add `-D_POSIX_C_SOURCE=200809L` to get the declarations. See man page for details. Here is an extract:

       fmemopen(), open_memstream(), open_wmemstream():
           Since glibc 2.10:
               _XOPEN_SOURCE >= 700 || _POSIX_C_SOURCE >= 200809L

It is also possible to add `#define _POSIX_C_SOURCE 200809L` to all files which use either `fmemopen` or `open_memstream`.

DanBloomberg:
Nice!

Just removing -ansi fixes the problem.

Is there any reason to add the #define _POSIX... as you suggest, or if I
do, is it OK to expose it to all platforms?

egorpugin:
It's better decide what standard to specify to C compiler.
As I wrote here https://github.com/DanBloomberg/leptonica/pull/116#issuecomment-177447476
Is it -std=gnu89 (in this case we should not fix anything) or -std=c89 (here we need to fix some errors and warnings like above)?
When we'll have explicit standard specification, it will be impossible to violate it accidentally.

stweil:
I don't expect that adding that definition would harm those platforms which don't need it. 

    #ifdef  HAVE_CONFIG_H
    #include "config_auto.h"
    #endif  /* HAVE_CONFIG_H */
    
    #if HAVE_FMEMOPEN && !defined(_POSIX_C_SOURCE)
    #define _POSIX_C_SOURCE 200809L
    #endif

egorpugin:
But still... Do you want to add a strict standard conformance option? Maybe c89? If you are ready to add posix workarounds.

DanBloomberg:
Egor, let me check with my mingw expert.  It might take a couple of days.

egorpugin:
Sure.
The choice is between 'c89' (doesn't compile at the moment) and 'gnu89' (compiles).

DanBloomberg:
Egor, Stefan,

This is what I've found out.

Compilation works with c89 and c99 if you undefine strict ansi:

   CC =            gcc -std=c89 -U__STRICT_ANSI__ -D__USE_MINGW_ANSI_STDIO
   CC =            gcc -std=c99 -U__STRICT_ANSI__ -D__USE_MINGW_ANSI_STDIO

(The mingw flag lets mingw find gnu extensions; it doesn't affect unix, and
likely won't affect windows.)

And as you already know, the code compiles with gnu89 and gnu99, because
those undefine strict ansi.

I don't have any opinion on what we should use, except that whatever we
decide should make it as easy as possible to write code that runs on all
the platforms (and not cause much disruption to the current code base).
Anyone have a strong opinion?

  -- Dan

PR #124, opened 2016-01-31
Clean handling of temporary files and fix compilation for Windows

stweil:
* Remove unneeded extern declaration for function fmemopen.

* Use existing function fopenWriteWinTempfile for fopenReadStream.

* Compile function fopenWriteWinTempfile only for Windows.

* Visual Studio does not provide the function mkstemp.
  Replace it by a combination of genTempFilename and _open.

* Replace the deprecated fdopen by _fdopen.

* Use "r+b" when calling _fdopen. It is sufficient there as it is not
  necessary to erase existing data in the file.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -194,6 +194,9 @@
 
 #ifdef _WIN32
 #include <windows.h>
+#include <fcntl.h>     /* _O_CREAT, ... */
+#include <io.h>        /* _open */
+#include <sys/stat.h>  /* _S_IREAD, _S_IWRITE */
 #else
 #include <sys/stat.h>  /* for stat, mkdir(2) */
 #include <sys/types.h>
@@ -1758,10 +1761,6 @@ convertOnBigEnd32(l_uint32  wordin)
 /*--------------------------------------------------------------------*
  *                        Opening file streams                        *
  *--------------------------------------------------------------------*/
-#if HAVE_FMEMOPEN
-extern FILE *fmemopen(void *data, size_t size, const char *mode);
-#endif  /* HAVE_FMEMOPEN */
-
 /*!
  *  fopenReadStream()
  *
@@ -1870,14 +1869,8 @@ FILE       *fp;
 #else  /* write to tmp file */
     L_WARNING("work-around: writing to a temp file\n", procName);
   #ifdef _WIN32
-    tmpdir = getenv("TMP");
-    snprintf(filename, sizeof(filename), "%s/%s", tmpdir ? tmpdir : ".",
-             fn_template);
-    handle = mkstemp(filename);
-    if ((fp = fdopen(handle, "r+b")) == NULL) {
-        L_ERROR("mkstemp stream not opened, %s\n", procName, strerror(errno));
-        return NULL;
-    }
+    if ((fp = fopenWriteWinTempfile()) == NULL)
+        return (FILE *)ERROR_PTR("tmpfile stream not opened", procName, NULL);
   #else
     if ((fp = tmpfile()) == NULL)
         return (FILE *)ERROR_PTR("tmpfile stream not opened", procName, NULL);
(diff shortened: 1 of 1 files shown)


egorpugin:
Still errors on win.

egorpugin:
As I remember you should use <io.h> header.

stweil:
Yes, io.h is needed for _open by Visual Studio. Obviously my test environment with MinGW gets it indirectly.

I added io.h now.

egorpugin:
Do you have VS installed? VS2015 Community is free for Open Source development.

stweil:
I cross compile on Debian GNU Linux ...

src/utils.c
```diff
@@ -1903,29 +1896,37 @@ FILE       *fp;
...
-char        filename[64];
-const char  fn_template[] = "mkstemp.XXXXXX";
-FILE       *fp;
+int  handle;
+FILE *fp;
+char *filename;
 
     PROCNAME("fopenWriteWinTempfile");
 
-    tmpdir = getenv("TMP");
```
egorpugin: Bad idea for windows to remove this.

src/utils.c
```diff
@@ -1903,29 +1896,37 @@ FILE       *fp;
...
-    snprintf(filename, sizeof(filename), "%s/%s", tmpdir ? tmpdir : ".",
-             fn_template);
-    handle = mkstemp(filename);
-    if ((fp = fdopen(handle, "w+b")) == NULL) {
-        L_ERROR("mkstemp stream not opened, %s\n", procName, strerror(errno));
+    if ((filename = genTempFilename("/tmp/", "lept", 1, 1)) == NULL) {
```
egorpugin: On windows there is no /tmp. You should take root dir as above. 'tmpdir = getenv("TMP");'

src/utils.c
stweil: Not necessarily. It is no longer needed here because the function now uses `genTempFilename`. That function replaces "/tmp/" by the Windows path for temporary files. I must admit that I don't like that kind of code, but it works.

src/utils.c
stweil: See comment above.

egorpugin: I see.

DanBloomberg: Need to restore the #ifdef _WIN32 to lines 1909 and 1927.

The reason is because I generate allheaders.h, and the function must be visible from all platforms (unix in my case) to put its declaration in allheaders.h.  From unix, it's just an empty function.

stweil: Patch updated to always include that function. I added a NULL return for the non Windows case, because otherwise compilers and code analyzers would complain about a missing return value.

DanBloomberg:
Stefan, all the 2400+ functions in allheaders.h are the externally visible api.

Some parts of the library are very stable (formatted image file I/O, basic image processing and analysis functions).   Other parts, such as more complicated applications like the text (digit) recognizer (recog*.c) will continue to evolve.   There is always a tradeoff between increasing functionality, simplifying the implementation, improving the consistency of the interface, and avoiding changes that will cause users to accommodate.

amitdo:
[_open](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/open-wopen?view=msvc-170)

>Opens a file. These functions are deprecated because more-secure versions are available; see [_sopen_s](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/sopen-s-wsopen-s?view=msvc-170)

DanBloomberg:
The only place I'm aware of using _open() is in fopenWriteWinTempfile(), in utils2.c, l.1993.
We also use _fdopen() in that same function.
Is this really a problem?  Are these microsoft functions posix and supported on other platforms?

I found this on the web:
https://www.anycodings.com/1questions/5140071/what-is-the-sopens-equivalent-for-the-open-function

amitdo:
Hi Dan,

First, I'm a Linux guy. not a Windows guy. I don't use Windows since 2004.

I don't know if there is a real safety issue with `_open()` (or with your usage of this function).

You are right. `_open()` is used only once.

https://github.com/DanBloomberg/leptonica/blob/0ffbc6822c23725b5b9f6876e2620a22ba3689f4/src/utils2.c#L1993

I saw this `_open()` function in the code, googled it and found the Microsoft docs.

[_open()](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/open-wopen?view=msvc-170) is Microsoft's alias for [open()](https://www.freebsd.org/cgi/man.cgi?sektion=2&query=open), which is part of POSIX 2008. By default, if you use the standard name instead of the alias, the MS compiler will print a warning. See [compatibility](https://learn.microsoft.com/en-us/cpp/c-runtime-library/compatibility?view=msvc-170).

`fdopen()` - POSIX 1988, `_fdopen()` - MS alias.

These underscore aliases for POSIX functions are done only by MS. Other libc implementations don't have them.

[_sopen_s()](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/sopen-s-wsopen-s?view=msvc-170) is non-standard. MS has other `_func_s()` 'safe' function for their own functions (like [_sopen()](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/sopen-wsopen?view=msvc-170)) and for standard C/POSIX functions.

>https://www.anycodings.com/1questions/5140071/what-is-the-sopens-equivalent-for-the-open-function

This link is a copy of:

https://stackoverflow.com/questions/6963659/what-is-the-sopen-s-equivalent-for-the-open-function

I can understand this 'politically motivated' remark.

DanBloomberg:
Thanks for the clarification.  I will never understand the engineering decisions of Microsoft.  For example, if they believe that _open() is not safe, why not improve the implementation instead of changing the signature?

I am even less of a windows programmer than you are, and I'd rather take the conservative route of leaving something alone that's not obviously broken.

amitdo:
Make sense to me :-)

PR #128, opened 2016-02-27
Fix use of fmemopen in src/utils.c

stweil:
src/utils.c uses the macro HAVE_FMEMOPEN which comes from config_auto.h,
so it must include that file.


DanBloomberg:
Thanks for fixing this.

PR #130, opened 2016-03-08
Using a simple typecast in tifio.c for successful compilation on VS2010

sumedhghaisas:


DanBloomberg:
Thanks for fixing this.

  -- Dan

PR #132, opened 2016-04-03
makefile.static: derive endianness without need to run binaries

rofl0r:
this makes it possible to use a cross-compiler.

### DIFF
--- a/src/makefile.static
+++ b/src/makefile.static
@@ -124,7 +124,7 @@ CC =		gcc -std=c99 -U__STRICT_ANSI__ -D__USE_MINGW_ANSI_STDIO
 #CC =		g++ -Wunused -D_BSD_SOURCE -fPIC
 
 #   Test for processor endianness (valid with gnu make only)
-ENDIANNESS := $(shell $(CC) -o endiantest endiantest.c; ./endiantest; rm -f endiantest)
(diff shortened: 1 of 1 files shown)


DanBloomberg:
This is very nice.  This works with the static makefile, but does not affect the other two ways of building the library.  You say it is possible to use a cross compiler (e.g., mingw).  How?

What happens with automake (Makefile.am)?  That distributes 4 different endianness.* files.

Or with cmake (CMakeLists.txt), which apparently has a dependency on endianness.h.

Is it possible to use this technique with these multiplatform utilities?

  -- Dan

rofl0r:
yes, it is possible to use a crosscompiler. my build process is this:
https://github.com/sabotage-linux/sabotage/blob/master/pkg/liblepton#L17
so basically you can just call
```
make -f makefile.static CC="arm-linux-musleabihf-gcc -isystem /path/to/sysroot/include -L /path/to/sysroot/lib"
```
and that's it.
(the sysroot is where the other crosscompiled libs and headers live.)
as for cmake, i try to avoid that when possible. the boilerplate needed to compile it is much higher: see this template file https://github.com/sabotage-linux/sabotage/blob/master/KEEP/pkg_skel/cmake#L8

PR #135, opened 2016-04-09
Correct minor deviances in function comments

pullmoll:
Change "Return" without colon to "Return:"


DanBloomberg:
Thanks.

Before I 'pull", can you change:

watershed.c,: leave line 240 as is.

All the *Destroy() functions should have a comment similar to that one.  For example, see box*Destroy() in boxbasic.c

PR #136, opened 2016-04-12
Doxygen: add a quick hack converter + Doxyfile

### LINKED ISSUE #126: Documentation online
Can you please publish recent documentation of leptonica?
Something like https://tpgit.github.io/UnOfficialLeptDocs/leptonica/ or https://tpgit.github.io/UnOfficialLeptDocs/leptonica/functions.html or doxygen generated (like tesseract-ocr.github.io)

pullmoll:
The converter relies on the rather constant formatting of
the current function's block comments and tries to turn
them into proper doxygen format.

The "/*!" doxygen comment introducer and the "\file" tag
are inserted when the first line with exact format
`" *  <basename>"` is found, e.g. `" *  adaptmap.c"`.

The included `Doxyfile` is a first example for a configuration.

This PR fixes minor deviances in comment format:

+ Add missing doxygen comment introducers "/*!".
+ Some function's comments had "Return" without a colon ("Return:")
+ Many function's comments used `@name` to reference parameters,
  while doxygen expects `%name` for references (`@` is a command introducer equal to `\`).
+ Add a new attribute `<inout>` for parameters in `*Destroy()` functions and
  unify the format of the comment.
+ Move "Input:" and "Return:" to column 8 where they were in another column
+ Replace "Note:" with "Notes:" where appropriate to make conv2doxy
  correctly detect the detailed description in function's block comments.
+ Move "Notes:" to column 4 where it was in another column.
+ Replace sub-paragraph " - " with " ~ " to avoid doxygen interpreting the
  text thereafter as a list entry; i.e. keep pre-formatted indentation.
+ Replace "null on error" with "NULL on error" in \return comments
+ Replace "Input: stream" with "Input: fp (file stream)" where appropriate.

The idea after merging this pull request is:

+ the source tree could or should be tagged with something like `Before transition to doxygen comment style`.
+ `conv2doxy.c` should be compiled and the binary be run once on every `src/*.c` file with e.g. `ls src/*.c|xargs ./conv2doxy`
+ now do something like `git add src/*.c`, `git commit -m "After transition to doxygen comment style"` and `git push`.
+ finally `doxygen` can be run and the resulting `doc/html` be published at some server.
+ the output `doxygen.log` can be used to identify more parameter name mismatches etc.
+ whenever new functions are added or existing documentation is fixed or improved, run `doxygen` again and upload the result. I for example used `rsync -avH --del doc/html user@example.com:~/public_html/misc`.
+ the original, backup files can be removed with `rm src/*.c~`

This closes #126


DanBloomberg:
Jürgen,

Tremendous job!  There are over 220K lines of code in the library alone..

I need some clarification on the next steps.  Let's go through them:


*The idea after merging this pull request is:*

** the source tree could or should be tagged with something like Before
transition to doxygen comment style.*
I looked at several of the source files in your patch.  After
hand-correcting some things (listed above) so that conv2doxy would work,
haven't you already run conv2doxy on the corrected source?  I.e., doesn't
this patch do the transition to doxygen formatted comments?

** conv2doxy.c should be compiled and the binary be run once on every
src/*.c file with e.g. ls src/*.c|xargs ./conv2doxy*
Same question.  Haven't you already run the converter?

** now do something like git add src/*.c, git commit -m "After transition
to doxygen comment style" and git push.*
Same question.

  -- Dan

pullmoll:
> I looked at several of the source files in your patch.  After
hand-correcting some things (listed above) so that conv2doxy would work,
haven't you already run conv2doxy on the corrected source?  I.e., doesn't
this patch do the transition to doxygen formatted comments?

No, I actually wanted to avoid the *big change* coming through a pull request. I thought it was nicer to have a transition commited by you, after taking a look at the modified src/*.c files and deciding, if they can be like that in the future.

This patch actually only *prepares* the source tree by aligning some things that the conv2doxy parsing would otherwise fail to handle. There may still be some edge cases, where the conversion doesn't produce fully correct output.

> ** conv2doxy.c should be compiled and the binary be run once on every
src/*.c file with e.g. ls src/*.c|xargs ./conv2doxy*
Same question.  Haven't you already run the converter?

I ran conv2doxy and then doxygen locally, for a testing purposes, and uploaded the results to http://misc.voidlinux.de/html/, but I did not put the massively modified source in this PR.

The intention was to first let you take a look at the huge amount of modifications that conv2doxy applies, perhaps criticize and ask me to correct this or that, and only then decide, if this is actually what you (or we?) want. The amount of changes in this PR is relatively small compared to the final result.

DanBloomberg:
Just saw this, after sending the last email on another thread.  I'll not
add anything to that thread.  But please look at it, so that you understand
what I am confused about, and the experiment that I want to do (run
conv2doxy on some of your converted files).

As I presently understand:

* The patch you sent (above, with pull request), with lots of changes,
including /*! \file ... added to every source file at the top, is the
"small" change before running conv2doxy.

My problem is that I can't imagine what more happens after conv2doxy runs
(even though I've looked at the code).

Apologies for being so obtuse.

  -- Dan

On Tue, Apr 12, 2016 at 2:20 PM, Jürgen Buchmüller <notifications@github.com

DanBloomberg:
and, btw, the documentation you created at voidlinux.de is incredible!

pullmoll:
Dan, the easiest way to test the change is to clone my version of leptonica elsewhere and switch to the doxygen branch:
```
mkdir temp
cd temp
git clone http://github.com/pullmoll/leptonica
cd leptonica
git checkout doxygen
```

I know there's some way to peek at a pull request, and I think it's even documented here on github.com somewhere (manual merging).

pullmoll:
The command line instructions, cut+paste from another PR
```
git checkout -b pullmoll-doxygen master
git pull https://github.com/pullmoll/leptonica.git doxygen
```
Later, if you want to merge this branch into master
```
git checkout master
git merge --no-ff pullmoll-doxygen
git push origin master
```

pullmoll:
Dan, you should perhaps read the comments at github.com, because changes to them are not sent by email AFAICT. To me it seems you are missing some of my comment's details when II modified the original text.

pullmoll:
Rebased this PR to master.

DanBloomberg:
You're right.  I should always read the comments on github.

This is very helpful.  I expect to be able to play with your PR and merge with the master today.

DanBloomberg:
This looks fine.  I need to make a small change in 6 files that were 'over-converted', but I'll do that separately.

DanBloomberg:
Jürgen,

I can live with the 'massive' changes to the documentation.  This is clearly a step forward.  I expect to do the conversion and push 'after doxygen' tomorrow, after some more testing.

pullmoll:
Dan, I'm glad to hear you agree to switch to doxygen style!
I intend to add some more comments to currently undocument functions or enumeration values.
You will most probably see some more pull requests from me after the transition.

DanBloomberg:
You suggested tagging the source tree with "Before transition to doxygen comment style".

How do I do that?  A commit with just a message and no file changes?

DanBloomberg:
Basic problem: I don't know what a tag is in this context.  Google search just shows how to use a windows Source Tree gui to add a tag.

pullmoll:
You can add a tag with a command like
```
git tag -a v1.73 -m 'Before transition to doxygen comment style'
```
That would be associated with the current v1.73 release tag https://github.com/DanBloomberg/leptonica/releases/tag/v1.73

You may of course use any name for the tag. A tag here is nothing but a name for a specific state of the repository. Perhaps you don't even want or need a tag, as the change is obvious from the version history anyway.

DanBloomberg:
The deed is done.  As you can see from the commit mesage, I had to patch things up a bit after conv2doxy, but it all compiles and succeeds on my local tests.  Nice that we only have to do this once  :-)

pullmoll:
Oh, I see there was a bug in conv2doxy: the last read `line` would have to be written to `fo`after the loop and you instead fixed it manually.. sorry for the inconvenience ;-P

From now on we can take care of what `doxygen.log` reports as \param inaccuracies or as missing descriptions etc.

Do you plan to publish the doxygen output somewhere?

DanBloomberg:
A small percentage of functions are so trivial that they have no parameter descriptions.  I'm OK with that.

I didn't run doxygen on the source files.  I'm still making small edits in the .h files (not yet merged).

Where do you suggest publishing the doxygen output?  Is there a canonical place on github?

pullmoll:
Dan, I don't know if github provides other places besides the wiki to store a HTML tree.
Perhaps http://leptonica.com/doc/ would be a good choice?

zdenop:
For publishing html on github there are "pages". You can create them in [web browser](https://pages.github.com/) or [manually](https://help.github.com/articles/creating-project-pages-manually/). We did it this way for [tesseract doxygen documentation](http://tesseract-ocr.github.io/).

PR #137, opened 2016-04-13
Fix endian detection across all build methods

chewi:
Unfortunately Leptonica has been broken on big endian systems when built with autotools because endianness.h was never used.

This commit ensures it is generated by Autotools, CMake, and the static Makefile, so it has therefore been included unconditionally in alltypes.h. This will break the native Visual Studio build but that
hasn't been maintained in a long time. For the static Makefile, the detection should now work under any Make implementation, not just GNU.

With the detected endianness now preserved in a header, it is no longer necessary to define L_LITTLE_ENDIAN or L_BIG_ENDIAN manually when building against Leptonica. This is a substantial improvement as forgetting to do so would have resulted in broken behaviour on big endian systems despite a complete lack of errors or warnings. Note that it will still respect your choice if you do decide to define these manually.

Building universal binaries for OS X should theoretically work with all build methods but this hasn't been tested. Feedback would be appreciated.


DanBloomberg:
James,

Thank you for this!  I'll merge it now.

I noticed that in src/makefile.static, line 123 still defined ENDIANNESS.  But this is no longer used, right?  (But ENDIANNESS is used in src/CMakeLists.txt ...)

chewi:
Actually it is still used [here](https://github.com/DanBloomberg/leptonica/blob/a75b2e176af72b6d63798538a577a8886cd890be/src/makefile.static#L348) and it no longer needs GNU Make as the := changed to just =. Glad I could help. I need to check something about the OS X universal binary stuff later but it's probably fine.

egorpugin:
Please, move these changes
https://github.com/DanBloomberg/leptonica/commit/fd252ce0a17561b74f8cc02726601e5be121ac58#diff-95e351a3805a1dafa85bf20b81d086e6R85
to
https://github.com/DanBloomberg/leptonica/blob/master/cmake/Configure.cmake#L78

And remove one of test_big_endian() calls.

chewi:
Can do. Any particular reason? I'm not as familiar with CMake as I am with Autotools.

egorpugin:
1) Duplicate calls. 2) All config stuff goes to cmake/Configure.cmake.

chewi:
Oh I see, didn't notice you already had the test there. I'll submit that shortly.

PR #141, opened 2016-04-16
Deduplicate endianness test in CMake

chewi:
The `WORDS_BIGENDIAN` macro is set by Autotools but we don't use it.

Please let @egorpugin confirm before merging this. I didn't move the other lines above `configure_file` because the `ENDIANNESS` and `APPLE_UNIVERSAL_BUILD` variables are used directly by that and nowhere else so separating them would be confusing.

### DIFF
--- a/cmake/Configure.cmake
+++ b/cmake/Configure.cmake
@@ -75,7 +75,7 @@ set(functions_list
 )
 check_functions(functions_list)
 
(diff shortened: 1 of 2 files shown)


egorpugin:
Thanks @chewi , I'll rework this slightly and commit it myself. For now closing this...

egorpugin:
Added in e8674e0554

PR #143, opened 2016-04-16
binreduce.c: fix doxygen warnings (doxygen.log)

pullmoll:


DanBloomberg:
Strange that it substituted a description of the lookup table around l. 375, rather than using the name of the function.

pullmoll:
The intended use of doxygen's `\brief` is to give a short description of the function. The fact that 2600+ lines `\brief functionName()` are in the source is because there was no way to automagically tell what a good description would be.

In the long term, perhaps whenever a function is touched, the `\brief` lines could be changed. The doxygen output would be even more useful that way.

DanBloomberg:
That seems reasonable.  If I get enthusiastic, I could change all the \brief lines in a FILE.

Wouldn't want to do 2600 one-line descriptions in one sitting  :-)

PR #145, opened 2016-04-16
classapp.c: fix doxygen warnings (doxygen.log)

pullmoll:


DanBloomberg:
Some interesting errors.

One problem change is line 541.  The function returns a struct Numaa  (typedef'd to NUMAA).  I am not consistent in the naming on the return line: sometimes I use the internal variable which is a shorthand for the type of object (e.g., naa is an abbreviation for numaa or Numaa.  I wish doxygen would let this pass as acceptable.  In this case, the change was made to "naa NUMAA", which doesn't really make sense.

My preference now would be to use Numaa, the actual name of the struct, if that is what is being returned.  I can make these changes over the next several months if you think that would solve this problem.

  -- Dan

pullmoll:
The doxygen `\return` command imposes no constraint on the text following at all. In contrast to `\param`, it does not expect a variable name or type.

The `naa NUMAA` in this PR is my bad. I'll try to avoid this type of comment in future fixes for doxygen warnings and keep the description as is.

DanBloomberg:
No problem.  Thinking again about naming of things explicitly returned, we should do what I have been doing; namely, just using the local variable name for that returned object.  If it's a struct, the name should give a hint as to what type it is.

I say a "hint" because you may have noticed that I try to use well-abbreviated names for variables.  So for example, I might call an instance of double number array (struct L_Dna) a "da".  Likewise, for an array of these, which is a L_Dnaa, I'd use "daa".

PR #152, opened 2016-04-17
bilateral.c: fix doxygen warning (doxygen.log)

pullmoll:


pullmoll:
The following doxygen commands are redundant and can be removed:
+ `\param none` for a function without parameters, i.e. `(void)`.
+ `\return void` for functions returning nothing

DanBloomberg:
Good to know.

PR #177, opened 2016-04-17
doxygen: fix remaining doxygen warnings

pullmoll:
Fix all remaining param warnings and conversion bugs.
Now 'doxygen.log' should contain only warnings for
missing documentation for functions, enums, variables etc.


pullmoll:
Note: I also typedef'd the allocator and deallocator function pointer types.
This simplifies the use of these types in the `struct PixMemoryManager` and allows `setPixMemoryManager()` parameters to be described for doxygen.

DanBloomberg:
Nice job!!

PR #181, opened 2016-04-21
Remove Autotools build of src/xtractprotos to fix #179

### LINKED ISSUE #179: Not able to compile leptonica due to lack of src/xtractprotos.c files
Dan,

I am not being able to compile leptonica since the removal of src/xtractprotos.c  from source (due to issue #119). 

make gives the follwoing output:
make[2]: *** No rule to make target `xtractprotos.o', needed by `xtractprotos'.  Stop.
make[2]: *** Waiting for unfinished jobs....
  CC     zlibmemstub.lo
make[2]: Leaving directory `/tmp/src/leptonica/src'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/tmp/src/leptonica'
make: *** [all] Error 2

Putting the file back from ./prog/xtractprotos.c fixed the error. Perhaps, to avoid the dupplication of the file, it should be enough to adapt the Makefile to compile from ../prog/xtractprotos.c

regards,
Guilherme

chewi:
The allheaders target now triggers a build of xtractprotos in the prog directory. I'm not sure if this is the proper way to do it but it's good enough to have the following work as expected.

```sh
./configure
cd src
make allheaders
```

I have assumed that allheaders.h should be written to the source directory rather than the build directory. Changing this would be difficult.

It doesn't work when building outside of the source tree but this was already broken. cpp, as invoked by xtractprotos, needs to locate endianness.h and this is written to the build directory. Perhaps xtractprotos should accept `-I` arguments or even pass all additional arguments to cpp.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,6 @@
 makefile
 Makefile
 Makefile.in
-xtractprotos
 /aclocal.m4
 /autom4te.cache
 /config/
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -73,7 +73,6 @@ pkginclude_HEADERS = allheaders.h alltypes.h                    \
  readbarcode.h recog.h regutils.h stack.h                       \
  stringcode.h sudoku.h watershed.h
 
-noinst_PROGRAMS = xtractprotos
 LDADD = liblept.la
 
 EXTRA_DIST = arrayaccess.h.vc                                   \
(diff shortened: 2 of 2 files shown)


chewi:
Just had a better idea regarding the last point. What if xtractprotos respected `${CPPFLAGS}` and passed it as arguments to cpp? That feels much cleaner.

DanBloomberg:
James,

Thank you for the fix!!

The answer to your first Q is yes, it should trigger building xtractprotos
in prog.  And yes, allheaders.h should live in the src directory.

As for your last point, I have no trouble accepting the current limitation
on building in the source directory.  We do that with the static makefile.
 (cmake does its own thing)

I'll just note that there is often a tension between (1) trying to make
something work in a more general situation and (2) the complexity of the
implementation.

(I usually come down on the side of simplicity, but that's because I put a
high priority on ease of implementation and esp. of maintenance.)

I'll merge your patch, and if you have second thoughts we can go from there
 :-)

  -- Dan

PR #182, opened 2016-04-23
Make float constants use float notation

pullmoll:
Shut some compilers' useless warnings about a "possible loss of data"
when converting a double constant (0.1) to a float. Append "f" (0.1f).


DanBloomberg:
Curious ... why did you delete this change?

pullmoll:
Hi Dan.
There are so many more float constants in the source that I thought it was a waste of time to modify just the few in this PR. It were just the `static const l_float32 = ...` lines.

PR #191, opened 2016-06-16
snprintf is available in VS2015

pnordhus:
Starting with Visual Studio 2015 `snprintf` is available and the macro is no longer needed.


DanBloomberg:
Philipp,

Thank you for the patch.  Good to see microsoft catching up with C89   ;-)

Issue #192, opened 2016-06-24
Include dependent Libraries


peirick:
For builds on windows its difficult to get all dependent libraries like libJpeg, libPng, libTiff, zlib. So maybe we can include the basic files of these libraries in this repository. The FreeImage library is already doing this. This should be done only as long as there is no nice way to manage this dependencies for windows builds.

DanBloomberg:
Tobias,

Stubs are provided for ALL functions in the image I/O files.  If the
libraries are not present, autobuild and cmake handle this, and the stubs
allow the library to build.

You can also manually force the stubs to be linked by modifying environ.h
and building the library with the static makefile.

peirick:
Hi Dan,
the problem I try to address is, that it is complicated to build leptonica on windows, that supports more than bmp files. You have to search for the current versions of the libraries and get them compiling and linking with leptonica. 
So It would be nice if you can run cmake on windows and it also generates the project files for libjpeg, libtiff, libzip. And you don't have to search the internet for current source code of that libraries.

DanBloomberg:
Tobias,

I think you're suggesting storing source files here, and having cmake
compile them as well as leptonica.  I don't know if it makes sense (really,
I have no idea).  Hadn't heard of it being done at the github level.

Those libraries are fundamental to any imaging application, whether it uses
leptonica or not -- there must be a more general way to get them.
Certainly we don't want to end up here:
      https://en.wikipedia.org/wiki/Dependency_hell
where leptonica has to worry about the version details of these libraries
(I ran into a problem with giblib, which you can read about at the
beginning of gifio.c.  It can be a real mess.)

As an alternative, is there a way to run a "script" file that will download
the basic I/O libraries and zlib to a windows machine, generate the project
files, and run them apart from leptonica?

egorpugin:
Storing other libs' sources in lept repo is not a right way.
But I have one possible solution. I'll check it and write it here.

peirick:
We could store the other libs' sources in another repository and load them, if needed, when cmake generates the solution.

egorpugin:
@peirick I've sent a PR. At the moment you try to build leptonica from my fork https://github.com/egorpugin/leptonica to see if workflow is smooth enough.
1. Clone it.
2. Download https://cppan.org/client/cppan-0.1.3-win32-client.zip Make sure you have vc2015 redist installed. 
3. Run cppan.exe from leptonica dir
4. Run cmake as usual.

[...]

PR #193, opened 2016-06-28
Add CPPAN support for windows builds

egorpugin:
Hi Dan,

I propose to add support of CPPAN (C++ Archive Network) for Windows build with dependencies. CPPAN is developed by me, now it's in alpha stage (but works pretty good). It's a C/C++ source package (packet) manager like CPAN (perl), CTAN (TeX) or CRAN (R-language). https://cppan.org/

Changes are quite small: one new file and couple of changes in CMakeLists.txt files that do not touch any existing build. CPPAN is activated only when 'cppan' command is invoked before build etc.

Such change is requested in #192.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 
 # build directories
 /build*
+/cppan*
 /win*
 
 # files in prog without a .
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,6 +75,18 @@ if (WIN32)
     endif()
 endif()
 
+if(EXISTS ${PROJECT_SOURCE_DIR}/cppan)
+    add_subdirectory(cppan)
+    add_definitions(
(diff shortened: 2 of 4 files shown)


DanBloomberg:
Egor, could lines 56 and 60 be pulled out of the #if/#else/#endif in
CMakeLists.txt?  Then you only need an #if USE_CPPAN section with 2 lines.

egorpugin:
Oops. A mistake! In 56 there should be no APPEND. Thanks!

DanBloomberg:
Egor,

Thanks for doing this.  I remember that the biggest support nightmare for leptonica on windows was linking these I/O libraries.

egorpugin:
We should probably create a Wiki page here on Github with build instructions.
Like this https://github.com/DanBloomberg/leptonica/wiki
I'll fill it tomorrow.

PR #201, opened 2016-08-24
fixed integer division by casting to float

vidiecan:


DanBloomberg:
Thanks for the report and the patch!

  -- Dan

vidiecan:
pleasure, thank *you* for the great library!

PR #203, opened 2016-08-26
fix build breaker

haikusw:
seems to fix the broken build that https://github.com/DanBloomberg/leptonica/commit/7ba075514b7231538bcedd2749a23bdb7b40e0a8 seems to have caused.


DanBloomberg:
thank you for fixing this!

  -- Dan

PR #206, opened 2016-09-01
In memory decoding of gif

peirick:
* usage of new function DGifOpen and custom input function
* only enabled if giflib version > 5.1. otherwise fallback to temp folder


DanBloomberg:
Tobias, thank you for this.  It looks sensible.

I am presently on vacation and unable to check it thoroughly.  Please give me about a week, and we can proceed.

peirick:
Nice. Thanks

PR #210, opened 2016-09-30
psio: Fix double free

stweil:
Coverity report:

CID 1365497 (#1 of 1): Double free (USE_AFTER_FREE)
18. double_free: Calling free frees pointer tempfile
which has already been freed.


DanBloomberg:
nice -- I forgot to remove that line 2 weeks ago when changing the way tempfiles are handled.

DanBloomberg:
should probably pull out the two braces (no longer needed), but I just merged it  :-)

PR #213, opened 2016-10-02
Fix some resource leaks (reported by Coverity Scan)

stweil:
Signed-off-by: Stefan Weil sw@weilnetz.de

### DIFF
--- a/src/adaptmap.c
+++ b/src/adaptmap.c
@@ -1274,10 +1274,11 @@ PIX       *pixm, *pixt1, *pixt2, *pixt3, *pixims;
 
         /* Finally, for each connected region corresponding to the
          * fg mask, reset all pixels to their average value. */
-    if (pixim && fgpixels) {
+    if (pixim && fgpixels)
         pixSmoothConnectedRegions(pixm, pixims, 2);
+
+    if (pixim)
         pixDestroy(&pixims);
-    }
 
     *ppixm = pixm;
     pixCopyResolution(*ppixm, pixs);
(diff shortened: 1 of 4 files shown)


DanBloomberg:
Stefan,

Attached is the way I'd do the adaptmap.c fix.

Note that in cleanup you don't need to test for null ptrs in the structs.
All the *Destroy() functions do that test.

The L_ERROR (etc) functions are convenient when you don't want to return
right away.  Note that the string for L_ERROR, L_WARNING and L_INFO should
end in a newline.

In blend.c, line 211 is not necessary (pixc is NULL).  I'd also handle the
other errors differently.  See attached blend.c.

There were actually 3 places in that function in binexpand.c where there
was potential loss.  However, all table builds should succeed, as they're
just very small mallocs and there are no input parameters that could cause
problems.  So it's best just to remove the tests.  (attached)

  -- Dan

stweil:
I'm sorry, but I did not find the attachments which you mentioned. The commits for `src/adaptmap.c` and `src/blend.c` are now modified as you suggested. Could you add your notes directly in the patch view for `src/binexpand.c`? That would be easier for me to see which parts of the code you are referring to.

src/adaptmap.c
```diff
@@ -2202,15 +2203,26 @@ PIXCMAP   *cmap;
     nar = numaGammaTRC(1.0, 0, L_MAX(1, 255 * rval / mapval));
     nag = numaGammaTRC(1.0, 0, L_MAX(1, 255 * gval / mapval));
     nab = numaGammaTRC(1.0, 0, L_MAX(1, 255 * bval / mapval));
-    if (!nar || !nag || !nab)
+    if (!nar || !nag || !nab) {
+        numaDestroy(&nar);
+        numaDestroy(&nag);
+        numaDestroy(&nab);
```
stweil: Example of patch comment: The preceding lines should be ok now.

src/blend.c
```diff
@@ -208,6 +208,7 @@ PIX       *pixc, *pixt, *pixd;
     boxDestroy(&box);
     if (!pixc) {
         L_WARNING("box doesn't overlap pix\n", procName);
+        pixDestroy(&pixt);
```
stweil: Here I removed an unneeded call of `pixDestroy(&pixc)`.

DanBloomberg:
Yes, I know it's not the canonical way to do github reviews, but there were
enough changes in 3 of the files so that it seemed easiest just to make
them and send you the file, so you can use them exactly as received in your
pull request.

Github may have dropped the attachments, so I will forward them to your
real email address.

PR #214, opened 2016-10-09
colorquant1: Remove wrong call of pixDestroy

stweil:
This fixes a new error reported by Coverity:

CID 1368189 (#1 of 1): Uninitialized pointer read (UNINIT)


DanBloomberg:
thanks!  I just introduced that bug!

PR #215, opened 2016-10-10
added missing include dir for templated sources

dschoerk:
LeptonicaConfig.cmake (in build dir) was missing the include reference to the "src" folder inside the build directory. In my case another project wasn't able to find "endianness.h" which resides in that folder.

This build request adds the missing include directory inside the build dir to the include path.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,7 +104,7 @@ include(Configure)
 
 configure_file(${AUTOCONFIG_SRC} ${AUTOCONFIG} @ONLY)
 
-set(INCLUDE_DIR ${CMAKE_SOURCE_DIR}/src)
+set(INCLUDE_DIR "${CMAKE_SOURCE_DIR}/src" "${CMAKE_CURRENT_BINARY_DIR}/src")
 
 configure_file(
     ${CMAKE_SOURCE_DIR}/cmake/templates/LeptonicaConfig-version.cmake.in


DanBloomberg:
Dominik, I'm not familiar with the syntax and structure of cmake templates, so I'll let Egor make the call on this.

But I have two questions.  endianness.h is in the source directory, along with the other include files, so I don't see why you need a second directory added to the list.  Also, what is the purpose for using double-quotes for those directory variables?

  -- Dan

dschoerk:
"endianness.h.in" is in the folder <source_dir>/src
cmake translates it into "endianness.h". "endianness.h" is then inside <build_dir>/src/

i believe the quotes are needed if the quoted path contains a space. but that might be incorrect, i'm not a cmake expert :)

dschoerk:
here is an example where the generated LeptonicaConfig.cmake causes the dependent tesseract-ocr build to fail because the include dir is missing. note the commented out repo which causes the tesseract compilation to fail.

SET LEPTONICA_GIT="https://github.com/dschoerk/leptonica.git"
SET TESSERACT_GIT="https://github.com/tesseract-ocr/tesseract.git"

:: once https://github.com/DanBloomberg/leptonica/pull/215 got accepted switch to official repo
:: SET LEPTONICA_GIT="https://github.com/DanBloomberg/leptonica.git"

SET LEPTONICA_DEST="test_libleptonica"
SET TESSERACT_DEST="test_libtesseract"

SET VSVERSION="Visual Studio 14 2015 Win64"

git clone %LEPTONICA_GIT% %LEPTONICA_DEST%
git clone %TESSERACT_GIT% %TESSERACT_DEST%

pushd %LEPTONICA_DEST%
mkdir build
pushd build
cmake -G %VSVERSION% ..
cmake --build . --config Release
cmake --build . --config Debug
popd
popd

pushd %TESSERACT_DEST%
mkdir build
pushd build
cmake -D CMAKE_PREFIX_PATH=$PWD/../../%LEPTONICA_DEST%/build -G %VSVERSION% ..
cmake --build . --config Release
cmake --build . --config Debug
popd
popd

DanBloomberg:
I see: endianness.h is not exported in the clone (it's in .gitignore).

Any idea why others haven't reported this problem?

Can you check if your patch works without the double quotes?  Of if not, using just the two exterior double quotes -- see, e.g., the set command in l. 89:
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W1")

egorpugin:
Hi, What do you want to achieve?
If you need to include tess or lept into your C++ project, you could use CPPAN. https://cppan.org
It's very easy. See https://github.com/cppan/tesseract_example

Regarding this PR...
I think it's ok to apply it.
No one reports about it because the use case is very limited.

PR #216, opened 2016-10-17
In memory encoding of gif with giflib >= 5.1

peirick:
It uses `EGifOpen` to create an in-memory GIF-file, and it uses the `ByteBuffer` datastructure to write to expanding memory.

The test `prog\gifio_reg.c` should remove the line `#if HAVE_FMEMOPEN` to test the new in-memory encoding.

Please let me know if this merge request is ok, or what changes needs to be made.

Thanks :-)


DanBloomberg:
Hey, this looks great!

I'll fix up the gifio_reg test.  That #IF_HAVE_FMEMOPEN will be removed, and the display for the test will be cleaned up.

PR #220, opened 2016-11-11
Handle $TMPDIR when located under /tmp

knowledgejunkie:
When adding the remainder of the cdir string to give the full value of pathout in genPathname(), the current code hardcodes a $TMPDIR string length of 4 characters (src/utils.c, lines 3004-5).

The result of this hardcoding is that the function will re-append any additional characters in the $TMPDIR string beyond "/tmp" into the pathout string when determining its full contents.

This causes failures when trying to open and write to files in $TMPDIR on Debian (and possibly other Linux distributions) that do not use plain /tmp as $TMPDIR for an unprivileged user. On Debian, a regular user's $TMPDIR is /tmp/username, not /tmp.

This commit aims to ensure that the full $TMPDIR string length is used and that only nested path entries below $TMPDIR (and not "/tmp") are appended when determining pathout.

Diagnosed using strace after failures running tesseract-ocr and gscan2pdf. No failures were observed when running as the superuser; strace confirmed the issue was broken pathnames and not r/w privileges.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -2986,23 +2986,24 @@ l_int32  dirlen, namelen, size;
         stringCopy(pathout, cdir, dirlen);
     } else {  /* in /tmp */
             /* Start with the temp dir */
+        l_int32 tmpdirlen;
 #ifdef _WIN32
-        l_int32 len;
         char tmpdir[MAX_PATH];
         GetTempPath(sizeof(tmpdir), tmpdir);  /* get the windows temp dir */
-        len = strlen(tmpdir);
-        if (len > 0 && tmpdir[len - 1] == '\\') {
-            tmpdir[len - 1] = '\0';  /* trim the trailing '\' */
+        tmpdirlen = strlen(tmpdir);
+        if (tmpdirlen > 0 && tmpdir[tmpdirlen - 1] == '\\') {
+            tmpdir[tmpdirlen - 1] = '\0';  /* trim the trailing '\' */
         }
 #else  /* unix */
         const char *tmpdir = getenv("TMPDIR");
         if (tmpdir == NULL) tmpdir = "/tmp";
+        tmpdirlen = strlen(tmpdir);
 #endif  /* _WIN32 */
-        stringCopy(pathout, tmpdir, strlen(tmpdir));
+        stringCopy(pathout, tmpdir, tmpdirlen);
 
-            /* Add the rest of cdir */
-        if (dirlen > 4)
-            stringCat(pathout, size, cdir + 4);
+       /* Add the rest of cdir */
+        if (dirlen > tmpdirlen)
+            stringCat(pathout, size, cdir + tmpdirlen);
     }
 
        /* Now handle %fname */


DanBloomberg:
Nick, thank you for your clean patch to the problem, and your fine and complete description of the problem and your fix!

chewi:
I'm afraid this has broken the tests on Gentoo when running under the package manager, where we set `TMPDIR` to `/var/tmp/portage/media-libs/leptonica-1.74/temp`. If I revert this then all the tests pass.

I am confused because I made changes to `genPathname` in 1.73 specifically to ensure that `TMPDIR` is respected and yet you say it breaks under Debian? The code is quite misleading as it appears to hardcode `/tmp` in a number of places but that is actually rewritten using `TMPDIR` here.

Admittedly I don't use Leptonica, I only package it. In coming up with a solution, we both need to ensure that it works for real usage as well as the tests. I'll have a closer and and try tesseract-ocr later.

DanBloomberg:
James, this is interesting.  The rewrite for windows is necessary so that all programs simply refer to "/tmp" and the right thing happens.  In this case, "/tmp" needed to be rewritten on unix as well (because they had redefined the temp directory).  The change was supposed to affect only unix, not windows.

I believe I see where the problem is  In line 3002, we copy tmpdirlen bytes, which was the length of tmpdir BEFORE the trailing slash was removed in the windows branch.  So if we go back and use strlen(tmpdir) as before, it should work.  Can you verify this?

  -- Dan

chewi:
That doesn't help as `tmpdirlen = strlen(tmpdir)` is the last line of the unix branch. It's effectively the same thing.

DanBloomberg:
No, just pull line 3000 out of the unix section and move it in front of line 3002.

DanBloomberg:
Oh, wait.  I thought you were running unix code.  Sorry.

Anyway, that line does need to be moved; otherwise the windows branch won't work properly.

DanBloomberg:
Big mystery!  The change 2 weeks ago with the introduction of tmpdirlen works on ubuntu linux.

DanBloomberg:
James, another mystery is that because your temp directory begins with '/var' and not '/tmp', that entire section that was changed would be skipped.

chewi:
I may be misunderstanding you but I don't think so, `dir` will begin `/tmp` on entering the function. As I said, this worked under 1.73. I'll try to figure this out ASAP.

chewi:
So Dave, seemingly independently, found that this change doesn't make sense. Now I've had time to look at it more carefully, I have to agree. I've installed Tesseract and it's working fine but strace shows that it isn't touching /tmp at all. @knowledgejunkie, what exactly did you do to trigger the problem?

knowledgejunkie:
I was running tesseract as part of the processing toolchain in a scanning script. My $TMPDIR was set to /tmp/nick (my local user), and tesseract (leptonica) would blow up when trying to read/write the temporary file. Running strace showed it leptonica to access paths under /tmp that would never exist (e.g. /tmp/nick/nick/lept/...), investigation of which resulted in the initial patch.

The location of $TMPDIR (and its permissions of 700) is set via my shell's profile.

DanBloomberg:
Nick, I believe we now have a version that will work for everyone.  I just put it on github.  I have revised text (not yet up) on usage of genPathname() that should make clear how to avoid the problem that you have had (with /tmp/nick/nick/...).  That text is:

 * Notes:
 *      (1) This function generates actual paths in the following ways:
 *            * from two sub-parts (e.g., a directory and a file name).
 *            * from a single path full path, placed in %dir, with
 *              %fname == NULL.
 *            * from the name of a file in the local directory placed in
 *              %fname, with %dir == NULL.
 *            * if in a "/tmp" directory and on windows, the windows
 *              temp directory is used.
 *      (2) On windows, if the root of %dir is '/tmp', this does a name
 *          translation:
 *             "/tmp"  ==>  <Temp> (windows)
 *          where <Temp> is the windows temp directory.
 *      (3) On unix, if the TMPDIR is a subdirectory of "/tmp", such
 *          as "/tmp/username", enter %dir as if it were "/tmp" (without
 *          the subdirectory).  The directory will automatically be
 *          rewritten using TMPDIR; e.g.
 *             "/tmp/lept/regout"  ==>  "/tmp/username/lept/regout"

chewi:
I fear we're laying hacks on top of hacks here but if it works, I'll roll with it. Here's my output. Do you see /tmp when you do this?

```
$ strace -e trace=file -f tesseract test.png stdout                             
execve("/usr/bin/tesseract", ["tesseract", "test.png", "stdout"], [/* 87 vars */]) = 0
access("/etc/ld.so.preload", R_OK)      = 0
open("/etc/ld.so.preload", O_RDONLY|O_CLOEXEC) = 3
open("/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib64/libtesseract.so.3", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib64/liblept.so.5", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/gcc/x86_64-pc-linux-gnu/5.4.0/libstdc++.so.6", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/gcc/x86_64-pc-linux-gnu/5.4.0/libgcc_s.so.1", O_RDONLY|O_CLOEXEC) = 3
open("/lib64/libc.so.6", O_RDONLY|O_CLOEXEC) = 3
open("/lib64/libpthread.so.0", O_RDONLY|O_CLOEXEC) = 3
open("/lib64/libm.so.6", O_RDONLY|O_CLOEXEC) = 3
open("/lib64/libz.so.1", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib64/libpng16.so.16", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib64/libjpeg.so.62", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib64/libgif.so.7", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib64/libtiff.so.5", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib64/libwebp.so.6", O_RDONLY|O_CLOEXEC) = 3
open("/usr/share/tessdata/eng.traineddata", O_RDONLY) = 3
open("/usr/share/tessdata/eng.traineddata", O_RDONLY) = 4
open("/usr/share/tessdata/eng.traineddata", O_RDONLY) = 4
open("/usr/share/tessdata/eng.traineddata", O_RDONLY) = 4
open("/usr/share/tessdata/eng.traineddata", O_RDONLY) = 4
open("/usr/share/tessdata/eng.traineddata", O_RDONLY) = 4
open("/usr/share/tessdata/eng.traineddata", O_RDONLY) = 4
open("test.png", O_RDONLY)              = 3
This is a lot of 12 point text to test the
ocr code and see if it works on all types
of file format.

The quick brown dog jumped over the
lazy fox. The quick brown dog jumped
over the lazy fox. The quick brown dog
jumped over the lazy fox. The quick
brown dog jumped over the lazy fox.

+++ exited with 0 +++
```

knowledgejunkie:
It might also be pertinent to note that I am passing tesseract Group4 TIFF files as input.

knowledgejunkie:
I noted the same leptonica errors during the failed attempt to process the TIFF file as in the tesseract issue here: https://github.com/tesseract-ocr/tesseract/issues/364.

That issue was shown to be due to insufficent user permissions when writing the temporary files. The issue I am describing here does seem to be related, but is due to leptonica looking in the wrong place for the files as a result of the value of $TMPDIR.

DanBloomberg:
In summary, the original problem was patched by Nick, but that approach is different from what we are asking people to do on windows, so it won't work for windows temp files.  Nick was giving the actual directory to genPathname: /tmp/nick.

The way to make it work on all platforms is to require the same directory input on both unix and windows, as in the notes pasted above.

DanBloomberg:
And I just pushed a final change to genPathname.

chewi:
I've managed to reproduce the issue with Tesseract now, after outputting to pdf. I wish I could say that Dan's changes have fixed it but while the tests pass, real usage is still broken on current master.

```
TMPDIR=/tmp/foo strace -e trace=file -f tesseract test.tiff out pdf
...
mkdir("/tmp/foo", 0777)                 = -1 EEXIST (File exists)
mkdir("/tmp/foo/lept", 0777)            = -1 EEXIST (File exists)
open("/tmp/foo/lept.usu67m", O_RDWR|O_CREAT|O_EXCL, 0600) = 5
open("/tmp/foo/foo/foo/lept.usu67m", O_RDWR|O_CREAT|O_TRUNC, 0666) = -1 ENOENT (No such file or directory)
Error in fopenWriteStream: stream not opened
Error in pixWrite: stream not opened
open("/tmp/foo/foo/lept.usu67m", O_RDONLY) = -1 ENOENT (No such file or directory)
open("lept.usu67m", O_RDONLY)           = -1 ENOENT (No such file or directory)
Error in fopenReadStream: file not found
Error in l_generateG4Data: stream not opened
unlink("/tmp/foo/lept.usu67m")          = 0
Error in pixGenerateCIData: g4 data not made
Error in l_generateCIDataForPdf: file test.tiff format is 8; unreadable
Error during processing.
+++ exited with 1 +++
```

I'll look closer at the code.

DanBloomberg:
James, please see if any calls into leptonica explicitly use "foo" or TMPDIR.

chewi:
You mean Tesseract? Of course it doesn't use foo and I get just the same effect if I use bar. It may use TMPDIR but why shouldn't it? As I suspected, it works if I set TMPDIR to be somewhere outside /tmp. I think this function is in a feedback loop.

I'm pretty sure I've said before that the whole concept seems flawed. The "temporary" patch in tesseract-ocr/tesseract#364 uses `mkstemp`, which is what we should be using. Windows has `_mktemp_s`. Why do you care what the temporary files are called? Predictable temporary filenames are actually considered a [security risk](https://www.owasp.org/index.php/Insecure_Temporary_File). Please give me some time to knock something together.

chewi:
Now I see why you care. It's for code such as this in dewarp4.c.

```c
        pathname = genPathname("/tmp/lept/dewarp1", buf);
        pixWrite(pathname, pixd, IFF_PNG);
        pixaDestroy(&pixa);
        pixDestroy(&pixt);
        pixDestroy(&pixd);
        LEPT_FREE(pathname);
    }
    bmfDestroy(&bmf);
    fprintf(stderr, "\n");

    fprintf(stderr, "Generating pdf of contour plots\n");
    convertFilesToPdf("/tmp/lept/dewarp1", "arrays_", 90, 1.0, L_FLATE_ENCODE,
                      0, "Disparity arrays", "/tmp/lept/disparity_arrays.pdf");
```

Ech. I need to think about this.

DanBloomberg:
James,

Yes, there's a bit of that for debugging.

(1) See l_makeTempFilename(), which uses mkstemp and may be of use for fixing this problem in tesseract code.

(2) I've now (finally) made it clear how to use genPathname(), and that you can't use TMPDIR as input to it. You have to write the temp file in a CANONICAL format, same on unix and windows, which is "/tmp/...".  Otherwise, it's a user error.

(3) re: "whole concept seems flawed".  Look at the library.  It is necessary to write temp files in known places, at least for regression tests and debugging. genPathname enables that. Because debug code is not used in production environments, I don't believe it should be considered a security issue.

chewi:
> 2) I've now (finally) made it clear how to use genPathname(), and that you can't use TMPDIR as input to it. You have to write the temp file in a CANONICAL format, same on unix and windows, which is "/tmp/...". Otherwise, it's a user error.

Okay but I'm struggling to understand exactly what the function is supposed to be for. It seems like it's partly to join an arbitrary directory and filename and partly to rewrite /tmp to the real temporary directory. These two operations can conflict. If you want to do the former but not the latter then who's to say that arbitrary directory won't start with /tmp? I think these should be two separate functions. If you have a function for explicitly returning a temporary file or directory then there's no need to do /tmp rewriting.

DanBloomberg:
OK, I'm going to write a little essay.  This is the best that I can do -- hopefully it will clarify the situation.

Just have a look in utils.h at all the places that genPathname() is used.  That should give you a good idea of why it exists.  For example, we always open files for reading and writing with fopenReadStream() and fopenWriteStream(), because that gives multi-platform support for reading and writing all files, including those in temp directories. (Joining directories and tailnames is just a tiny part of its usefulness.) After you've finished, read on ...

Hint: if there were no rewriting issue, then you don't need this function at all.  If both windows and unix always wrote to the temporary directory "/tmp", then you don't need it for much.

But the world is not that simple, and (1) tmp files are required for some applications and tests and (2) this should work on multiple platforms.  Leptonica would be seriously crippled if we couldn't write debugging files to places that don't matter much, and it would be buggy and difficult to maintain if we couldn't do regression tests.

As you'll remember from looking at utils.c, many of the uses for genPathname() are for generating and removing temporary directories, and for writing and reading files into those directories.

The rewriting that is done is the absolute minimum that I could get away with.  (As you may remember, we used to rewrite from "/tmp" to subdirectories of /tmp, which was confusing.)  Consider how this has evolved over the past 5 years, and the fact that we're now supporting over 200 source files in the library and over 250 programs in prog/, including 84 tests in 'make check'.

DanBloomberg:
split utils.c --> utils1.c and utils2.c

Added 3 "faq" type items to the comments at the beginning of utils2.c, that summarize the verbiage above.

chewi:
You asked earlier whether Tesseract makes any calls into Leptonica using `/tmp`. If you meant direct `genPathname` calls then there aren't any. Neither are there any in jbig2enc, the only other open source project directly using Leptonica that I'm aware of. If you meant indirect `genPathname` calls then you can hardly blame them as they may not have even looked at the description of `genPathname`.

chewi:
I believe this has now been fixed as of 1.74.0.

DanBloomberg:
Yes, the issue is resolved.

DanBloomberg:
@chewi 
@jbreiden

Well, not yet resolved.  We have a Debian bug in tesseract due to the rewriting from genPathname().

If TMPDIR = /tmp/any/path and you call genPathname("/tmp/any/path", foo.x), you get:
      /tmp/any/path/any/path/foo.x   !!
Also, if you're just looking for /tmp/foo.x, this will rewrite to /tmp/any/path/foo.x, and you won't find it.

So here's the current plan:
 for unix, we don't rewrite.  We ignore TMPDIR.  It actually makes no sense to do the rewrite because /tmp exists in unix.  Only do the rewrite for Windows.

Are we going to have issues with this limited rewrite policy on other platforms (e.g., mingw)?

jbreiden:
The Debian bug is https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=885704 

I am the Debian package maintainer, and am going to patch the Debian package to completely remove path rewriting. I don't know if this will break something  (tests, debug paths, or even some normal code paths) but I think we are going to find out.

chewi:
I don't like the idea of not respecting `TMPDIR`. It's bad manners at best but at least you've kept it for the test suite, which as a package maintainer who doesn't actually use Leptonica, is all I have time to care about by this point. I still stand by what I said earlier, that the concept is flawed. Sorry, but you haven't managed to convince me otherwise.

jbreiden:
Let's be clear about this. @dbloomberg is upstream, @jbreiden and @chewi are package maintainers for two different Linux distributions. I personally think the best overall approach is this:

(1) Get rid of path rewriting everywhere.

(2) Any temporary files produced by a production code path should use `mkstemp` or similar. There was already an exercise last summer for this. Hopefully didn't miss anything. These will respect `TMPDIR` automatically.

(3) Temporary files produced by tests or debug code paths should have predictable names. Otherwise they are not so useful. To get that, we should use a utility function. So instead of seeing things like `"/tmp/lept/dewarp1"` in debug sections of code, use something like `debugPath("lept/dewarp1")`. The utility function should respect TMPDIR. It should also understand weird operating systems like win32.

Upstream is currently think about something else (just remove rewriting on non-win32). @chewi, can you say exactly what you think about upstream's idea, and also what you think about the larger proposal I just wrote above? I wasn't totally sure who and what you were responding to in your most recent comment.

DanBloomberg:
@chewi 

You can see from the two examples I gave above that the current implementation is busted on unix. The second example shows the problem we get into when respecting TMPDIR on a system where you can also write directly to /tmp.  So it has to be fixed.

I don't see any alternative, other than only using temp paths generated by a specific function, and doing this throughout leptonica and the progs, which would be a huge ordeal, which "upstream" cannot imagine doing.

Are you aware of any problems that you expect from this change for the leptonica regression tests on gentoo?

chewi:
Sorry, I was mainly responding to Dan with regards to commit DanBloomberg/leptonica@c72049d16d36ca21d8549f9ae8eefc4fdf5a1807. This commit makes things just about good enough as far as I'm concerned. Users installing the package with tests enabled won't unexpectedly get tons of junk left in /tmp (our package manager sets `TMPDIR`). It should also behave at runtime, short of not respecting `TMPDIR`, which isn't ideal but if anyone complains, I can just tell them to take it up with upstream.

@jbreiden, I agree with everything you've said and I've said as much myself in the past but as Dan says, it may not be that simple. I'm not so sure myself but I should put the code where my mouth is. I'm a bit tied up right now unfortunately.

jbreiden:
@chewi With commit c72049d, I think running tests will spam /tmp no matter how `TMPDIR` is set. Debian does care because Debian doesn't run the tests. Sounds like Gentoo is a different story.

chewi:
Ah, you're right, I misread it. I can disable the running of tests in the package but some users like the extra assurance that it will actually work. It is also helpful for me as we have seen the odd breakage arising from different architectures and library versions.

Leaving junk in /tmp is mildly annoying at the best of times but in this case, it dumps about 300MB of stuff in there and that really sucks if your /tmp is in RAM.

DanBloomberg:
@chewi @jbreiden 

Disable the tests? Let's be clear on this: we were careful to make sure that the tests put the temporary files all neatly gathered in subdirectories of /tmp/lept, so it's trivial to remove them after running the test:
     rm -rf /tmp/lept/*

Rhetorical question: How can this be even _mildly_ annoying?

chewi:
It is annoying if you don't even know they're there but you're wondering where a chunk of RAM went. Some Gentoo users like to enable tests to ensure that their software is going to work properly. They're not going to hack Leptonica, they may not even know what it is, it'll just get pulled in as a dependency of Tesseract.

I could add the `rm` as a workaround, I did consider it before, but at least as of 1.74.4, I get left with over 200 files like `lept.pPR0t7` sitting outside the `lept` subdirectory. Of course, I could make it `rm -rf /tmp/lept /tmp/lept.*` but is it supposed to do that?

DanBloomberg:
James, you answered my 'rhetorical' question  :-)

I don't know where those files are coming from.  They look like system generated temporary files.  When I run a full set of 102 tests in ubuntu, I get this in /tmp/lept/:

drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 adapt/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 alpha/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 baseline/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 boxa/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 class/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 class2/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 color/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 comp/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 conn/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 disp/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 encfonts/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 format/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 gif/
drwxr-xr-x 2 dbloomberg eng 69632 Jan  3 15:50 golden/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 pageseg/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 pdfseg/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 plots/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 pnm/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 psio/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 quad/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 rank/
drwxr-xr-x 2 dbloomberg eng 69632 Jan  3 15:50 regout/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 strfonts/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:49 tiff/
drwxr-xr-x 2 dbloomberg eng  4096 Jan  3 15:50 webp/

and nothing in /tmp itself.

chewi:
Maybe that was a quirk in 1.74.4 as I'm not seeing those extra files more. I do get a `/tmp/smooth` directory outside `/tmp/lept` though.

DanBloomberg:
Yeah, I can believe there are a couple of stray directories left that haven't yet been put under /tmp/lept/

amitdo:
The handling of temp files in leptonica is a never ending story...

jbreiden:
This command hunts down all hardcoded `/tmp` paths in Leptonica and abstracts them. Unforunately that's not a practical answer in C,  because you have free the memory used in the new pathname.

```
find -type f | xargs -n 1 sed -i -e 's/\"\/tmp\/\([^\"]*\"\)/debugPath(\"\1)/g'
```

There are 225 spots, here is one of them.

```diff
-        pixWrite("/tmp/lept/segment/colorseg2.png", pixd, IFF_PNG);
+        pixWrite(debugPath("lept/segment/colorseg2.png"), pixd, IFF_PNG);
```

DanBloomberg:
We've been over this.  Today, essentially all temp files are in a subdirectory of /tmp/lept/.

As I said, the few remaining ones that are not in some such subdirectory will be eliminated in exponential fashion (eg, with a half-life).

With 1.75.0, we have eliminated all temp file rewrites on unix.  If you are on unix and ask for /tmp/foo/...,, that's what you get.  If you ask for $TMPDIR/foo/..., you get that.  No rewrites, period.

PR #228, opened 2017-01-03
configure: Support pkg-config

stweil:
The following packages provide compiler and linker flags via pkg-config:

* zlib
* libpng
* libjpeg
* libtiff-4
* libwebp
* libopenjp2

Add the flags from pkg-config to the preprocessor flags.
Add code to include openjpeg.h if LIBJP2K_HEADER was not set.

Packages without a configuration for pkg-config work as before
with the old detection code.


stweil:
Both AppVeyor builds stopped after an 60 minutes. This seems to be unrelated to my pull request, but to be sure I restart the CI tests by closing and reopening this PR.

DanBloomberg:
Stefan,

The 60 minute timeout was an appveyor problem, not yours.  It happened to me yesterday.

Where does this pull request stand now -- is it safe?

stweil:
I think it is safe. Builds with `cmake` should not be affected at all. I have run tests with `./autobuild && ./configure && make` with Linux native, Windows cross and Windows native builds.

DanBloomberg:
thanks -- fingers crossed  ;-)

PR #230, opened 2017-01-03
binarize: Fix use of unitialized value

stweil:
Coverity scan report:

CID 1365363: Uninitialized pointer read

pixd is only assigned a value if ppixd != NULL.


DanBloomberg:
thanks!

PR #241, opened 2017-04-07
Improved png memory I/O performance.

tdhintz:
This primarily benefits Tesseract OCR on the Windows platform because it replaces file I/O with memory operations.

### DIFF
--- a/src/memio.c
+++ b/src/memio.c
@@ -1,0 +1,181 @@
+/*====================================================================*
+-  Copyright (C) 2017 Milner Technologies, Inc.  This content is a
+-  component of leptonica and is provided under the terms of the
+-  leptonica license.
+-
+-  Redistribution and use in source and binary forms, with or without
+-  modification, are permitted provided that the following conditions
+-  are met:
+-  1. Redistributions of source code must retain the above copyright
+-     notice, this list of conditions and the following disclaimer.
+-  2. Redistributions in binary form must reproduce the leptonica
+-     copyright notice, this list of conditions and the following
+-     disclaimer in the documentation and/or other materials
+-     provided with the distribution.
+-
+-  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+-  ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+-  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+-  A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL ANY
+-  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+-  EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+-  PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+-  PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+-  OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+-  NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+-  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*====================================================================*/
+
+/*!
+* \file memio.c
+* <pre>
+* libpng read/write callback replacements for performing memory I/O.
+* </pre>
+*/#include <string.h>
+#include <assert.h>
+#include "allheaders.h"
+#include "memio.h"
+
+/* 
+* <pre>Buffer allocation size.  Set this small enough to avoid consuming unnecessary memory,
+* but larger than an average image. </pre>
+*/
+#define BUFFER_SIZE 8192
+
+/*
+* <pre>
+* libpng callback for writing an image into a linked list of memory buffers.
+* </pre>
+*/
+void
+memio_png_write_data(png_structp png_ptr, png_bytep data, png_size_t len)
+{
+	MemIOData* thing;
+	PROCNAME("memio_png_write_data");
+
+	thing = (MemIOData*)png_get_io_ptr(png_ptr);
+	if (thing->m_Last->m_Buffer == 0)
+	{
+		if (len > BUFFER_SIZE)
+		{
+			thing->m_Last->m_Buffer = new char[len];
+			memcpy(thing->m_Last->m_Buffer, data, len);
+			thing->m_Last->m_Size = thing->m_Last->m_Count = len;
+			return;
+		}
+
+		thing->m_Last->m_Buffer = new char[BUFFER_SIZE];
+		thing->m_Last->m_Size = BUFFER_SIZE;
+	}
+
+	int written = 0;
+	while (written < len)
+	{
+		if (thing->m_Last->m_Count == thing->m_Last->m_Size)
+		{
+			MemIOData* next = new MemIOData;
+			next->m_Buffer = 0;
+			next->m_Count = 0;
+			next->m_Size = 0;
+			next->m_Next = 0;
+			next->m_Last = next;
+
+			thing->m_Last->m_Next = next;
+			thing->m_Last = next;
+
+			thing->m_Last->m_Buffer = new char[BUFFER_SIZE];
+			thing->m_Last->m_Size = BUFFER_SIZE;
+			thing->m_Last->m_Count = 0;
+		}
+			
+		int remainingSpace = thing->m_Last->m_Size - thing->m_Last->m_Count;
+		int remainingToWrite = len - written;
+		if (remainingSpace < remainingToWrite)
+		{
+			memcpy(thing->m_Last->m_Buffer + thing->m_Last->m_Count, data + written, remainingSpace);
+			written += remainingSpace;
+
+			thing->m_Last->m_Count += remainingSpace;
+		}
+		else
+		{
+			memcpy(thing->m_Last->m_Buffer + thing->m_Last->m_Count, data + written, remainingToWrite);
+			written += remainingToWrite;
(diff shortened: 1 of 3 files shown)


src/memio.h
```diff
@@ -0,0 +1,78 @@
...
+/*!
+* \file memio.h
+* <pre>
+* libpng read/write callback replacements for performing memory I/O.
+* </pre>
+*/
+#ifndef  LEPTONICA_MEMIO_H
+#define  LEPTONICA_MEMIO_H
+
+#include <png.h>
```
amitdo: You should add here
```
#ifdef  HAVE_CONFIG_H`
#include "config_auto.h"
#endif  /* HAVE_CONFIG_H */
```

and wrap the code ih
```
/* --------------------------------------------*/
#if  HAVE_LIBPNG 

/* --------------------------------------------*/
#endif  /* HAVE_LIBPNG */
/* --------------------------------------------*/
```
https://github.com/DanBloomberg/leptonica/blob/8a839a254e8d2e/src/pngio.c#L113

amitdo:
Travis failes. There are compilation errors.
https://travis-ci.org/DanBloomberg/leptonica/builds/219636596

src/memio.c
```diff
@@ -0,0 +1,181 @@
...
+-  NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+-  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*====================================================================*/
+
+/*!
+* \file memio.c
+* <pre>
+* libpng read/write callback replacements for performing memory I/O.
+* </pre>
+*/#include <string.h>
```
amitdo: First, put a linebreak between the `/*` and the includes.

Second, in the rest of the codebase `#include "allheaders.h"` comes after the cstdlib headers:
https://github.com/DanBloomberg/leptonica/blob/8e7190ccfb46c71a36b95afc9ed402704e476309/src/pix4.c

tdhintz:
To be clear, someone with experience with the build environment needs to make all of these changes.

Sent from my iPad

On Apr 7, 2017, at 8:05 AM, Amit D. <notifications@github.com<mailto:notifications@github.com>> wrote:


@amitdo commented on this pull request.

________________________________

In src/memio.c<https://protect-us.mimecast.com/s/oXoMBgSm6QwIQ?domain=github.com>:

> +-  A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL ANY
+-  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+-  EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+-  PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+-  PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+-  OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+-  NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+-  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*====================================================================*/
+
+/*!
+* \file memio.c
+* <pre>
+* libpng read/write callback replacements for performing memory I/O.
+* </pre>
+*/#include <string.h>


First, put a linebreak between the /* and the includes.

Second, in the rest of the codebase #include "allheaders.h" comes after the cstdlib headers:
https://github.com/DanBloomberg/leptonica/blob/8e7190ccfb46c71a36b95afc9ed402704e476309/src/pix4.c<https://protect-us.mimecast.com/s/ANO9BvFVRXWUd?domain=github.com>

amitdo:
This PR is supposed to solve this Tesseract issue: 
https://github.com/tesseract-ocr/tesseract/issues/522

tdhintz:
With this change I'm seeing about a 8 - 25% performance boost in Tesseract 4.0.0 LSTM w/OpenMP when passing a bi-tonal bitmap on Windows x64.  Visual Studio 2017 was used for the builds.  The higher number occurs when concurrency is used to process multiple pages at the same time.  The lower number is experienced when performing 1 page OCR and no concurrency.  I assume your performance will depend significantly on the number of png images generated as artifacts of the OCR process.  I predict the more artifacts that are generated the greater the savings.

I notice that the code may be leaking the contents of "state" in error conditions.

DanBloomberg:
8% improvement of the overall time on a single page seems surprising to me.

Can you break this down in terms of read time (both directly to memory and via file) and processing time,for a typical page of text at some stated resolution?

tdhintz:
I understand.  I posted because I'm interested in the results of others.  I'm running small samples so a little variation can skew a lot.  I have here a single page test that just averaged 13.08 over 3 runs without the change, and 3 runs with the change averaged 12.61 seconds.  This is a gross measurement of overall time using the Weld .Net wrapper and iterating characters. 4% if I did my math correctly for this sample.  Not 8% on that run but consistently measurable. 

I haven't isolated the measurements into a test harness so I can't break it down by where the time is gained. I seems logical that there should be some gain on Windows given the old implementation.  I see leptonica tries to give Windows a hint that the file doesn't actually need to be written but clearly there must be overhead using the temp directory even if the file isn't flushed (directory reads, security checks, file name generation, etc).

On systems with fmemopen() I'd have to guess that the two implementations are close to a wash but I have no way of measuring that.

DanBloomberg:
No question in my mind that for reading and writing via memory, doing it directly is better than via a temporary file and using a file stream interface -- even if it is done securely using tmpfile() or similar methods.  This should ultimately be enabled for all platforms.

Do you have any idea why Microsoft hasn't yet implemented the fmemopen and openmemstream() library functions, which are POSIX 2008?  How hard could this be...

tdhintz:
I don't have special insight, but Microsoft hasn't historically had a POSIX focus.  Maybe their new cross platform strategies will lead them there.

src/pngio.c
```diff
@@ -475,6 +481,363 @@ PIXCMAP     *cmap;
...
+	png_uint_32  xres, yres;
+	png_bytep    rowptr, trans;
+	png_bytep   *row_pointers;
+	png_structp  png_ptr;
+	png_infop    info_ptr, end_info;
+	png_colorp   palette;
+	png_textp    text_ptr;  /* ptr to text_chunk */
+	PIX         *pix, *pixt;
+	PIXCMAP     *cmap;
+	struct MemIOData state;
```
stweil: I suggest to remote `struct` here and simply use `MemIOData`.

src/memio.h
```diff
@@ -0,0 +1,78 @@
...
+
+	/*
+	* <pre>The allocated size of m_Buffer.</pre>
+	*/
+	int m_Size;
+
+	/*
+	* <pre>Pointer to the next node in the list.  Zero if this is the last node.</pre>
+	*/
+	MemIOData* m_Next;
```
stweil: Here `struct` is needed because `MemIOData` without `struct` is still undefined.

src/memio.h
```diff
@@ -0,0 +1,78 @@
...
+
+	/*
+	* <pre>Pointer to the last node in the linked list.  The last node is where new
+	* content is written.</pre>
+	*/
+	MemIOData* m_Last;
```
stweil: Here `struct` is needed, see above.

src/pngio.c
```diff
@@ -1213,6 +1576,353 @@ char        *text;
...
+	png_uint_32  xres, yres;
+	png_bytep   *row_pointers;
+	png_bytep    rowbuffer;
+	png_structp  png_ptr;
+	png_infop    info_ptr;
+	png_colorp   palette;
+	PIX         *pixt;
+	PIXCMAP     *cmap;
+	char        *text;
+	struct MemIOData state;
```
stweil: Remove `struct` here.

src/memio.c
```diff
@@ -0,0 +1,181 @@
+/*====================================================================*
+-  Copyright (C) 2017 Milner Technologies, Inc.  This content is a
+-  component of leptonica and is provided under the terms of the
```
stweil: I suggest to use the upper case form `Leptonica` here and also in the next line.

stweil:
> The Travis build failed.

Those two build errors are caused by missing `struct` keywords (see my inline comments).
Obviously the MS compiler tolerates non standard C code whereas gcc raises errors.

DanBloomberg:
Note also that the code must be strict ANSI C.

* No 'new', 'delete'
* all variable declarations immediately following an open brace

And there are some strict guidelines, in particular no asserts  (library cannot exit)

See also style-guide.txt in the root directory, mostly for formatting.

stweil:
> Note also that the code must be strict ANSI C.

@tdhintz, there are compiler options for Visual C which disable the Microsoft extensions. Try `/Za`, maybe in combination with `/Zc` or other options.

tdhintz:
The project used C++.  I don't recall whether I did that or the cmake.  I can't think in C so the prototype was C++ and I translated it back to C.  Performance with the new vs malloc() was better in a brief test so I left it that way for safety.  It is most probable that the difference was due to some factor other than the use of new, but I stuck with the lucky hat.  All the suggestions on code format and C compatibility have been diverting but the better question is whether the code improves performance on Win platform.  If it doesn't then this exercise has no benefit.

DanBloomberg:
You mentioned that the memio buffers can have a memory leak with a low-level failure (e.g., causing setjmp).  What is the current situation?

tdhintz:
The current commit handles memory correctly on error conditions.

DanBloomberg:
Let me take a day or two to verify that this will work OK with 'make allheaders', which generates allheaders.h, and that it passes all the regression tests for png I/O -- and I'll merge it as is.

After that, I'll need to incorporate memio.{h,c} directly into pngio.c to make its functions static, so that they do not get scooped up into allheaders.h by xtractprotos.  And I'll be able to rewrite the now-redundant *Stream* functions as thin shims of your Mem functions.

DanBloomberg:
Slightly modified version is in and working.

As it turned out, I couldn't easily make the pix read/write stream functions as shims of the mem functions, because the pixa read/write functions are now using the stream interface for each individual pix, and the mem functions are not designed to work consecutively.  It's possible to handle this in pixaReadMem(), for example, but it's not clear that it's worth the effort.

PR #243, opened 2017-04-08
Update adaptmap.c

lcarrere:
Check allocation of a bitmap in the function pixApplyInvBackgroundGrayMap


DanBloomberg:
Just curious.  Did you have a situation where the allocation failed?

lcarrere:
Hello Dan. Yes I was able in situation of high memory usage, especially on 32-bit process. Just let if know if you need any other information.

DanBloomberg:
Thanks for the suggestion.  Added this, plus two others in the same file.

PR #245, opened 2017-04-18
Disable libtiff warnings for best performance.

tdhintz:
libtiff tends to complain a great deal unnecessarily.  For example, TIFF allows vendor-defined tags but libtiff complains about them  The proposed change silences warnings for best performance.  libtiff errors are allowed to be reported.

### DIFF
--- a/src/tiffio.c
+++ b/src/tiffio.c
@@ -1132,6 +1132,11 @@ l_uint32   uval, uval2;
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


DanBloomberg:
Thanks!  I've slightly reformatted (for C) in a local client and will push the change.

DanBloomberg:
it's in

PR #254, opened 2017-05-31
do not call boxaSort with empty boxa

vidiecan:
do not emit a warning if no baseline is detected hence empty boxa2


DanBloomberg:
Thank you for your patch.

Looking at the issue, I decided a more extensive revision is warranted.  Namely, if we find there are no components, we should exit the function at that point, and the various destructors should be moved up as far as possible to simplify the exit.

Issue #253, opened 2017-05-30
autotools vs. cmake - different soname


amitdo:
`liblept.so` vs `libleptonica.so`

That's not good...

DanBloomberg:
Assuming that we will continue to use both autotools and cmake, Jeff points out
that it will be a considerable amount of work to change the debian soname (liblept.so).

How much trouble is it to change the cmake soname to agree with this?

egorpugin:
It's easy to change name in cmake.

DanBloomberg:
Will that be sufficient?  i.e.,  without changing the autotools name, which is defined in src/Makefile.am?

amitdo:
Why not in all platforms?

egorpugin:
`leptonica` is more beautiful than `lept`.
To keep compatibility on linux we'll rename to `lept` in cmake, but keep `leptonica` on win.
Or even keep it `leptonica`. Why to change? People do not use cmake build for creating debian packages.
And cmake users can live with `libleptonica`.

amitdo:
Many people, on all platforms, build Leptonica and Tesseract from source.

https://github.com/tesseract-ocr/tesseract/issues/957
Here someone build Leptonica using CMake and Tesseract using autotools.

zdenop:
I prefer compatibility before beauty ;-)

DanBloomberg:
@jbreiden

jbreiden:
zdenko is 100% right. We have enough chaos and confusion in life. Leptonica gets exactly one name for its library, and the author (Dan Bloomberg) gets to choose it, and he is chose liblept.so long ago. We stick with it because that is what he wants, and also because it has millions of installed users. Modify cmake build to match. Not negotiable.

stweil:
As long as other packages use `pkg-config` to get the library name, the name does not matter at all (assumed that `lept.pc` is correct). Currently `cmake` creates a broken `lept.pc` file which includes `Libs: -L${libdir} -lleptonica_OUTPUT_NAME-NOTFOUND`. Of course that does not work.

egorpugin:
Wait, wait. Who uses that cmake build?
Package maintainers keep using autotools? Well then, cmake won't touch their stuff.
People who use cmake+autotools build? It's crazy. Just don't do this. Don't mix them.
(Why people use cmake? Because it's simpler.)

[...]

PR #279, opened 2017-09-05
Fix cmake Release optimization level

Sineaggi:
Similar to [tesseract#1100](https://github.com/tesseract-ocr/tesseract/pull/1100), the cmake file installs an incorrectly named library. Changes `libleptonica` to `liblept`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -14,6 +14,8 @@ if (MSVC)
     set_source_files_properties(${src} PROPERTIES LANGUAGE CXX)
 endif()
 
+string(REPLACE "-O3" "-O2" CMAKE_C_FLAGS_RELEASE ${CMAKE_C_FLAGS_RELEASE})
+
 add_library                     (leptonica ${LIBRARY_TYPE} ${src} ${hdr})
 set_target_properties           (leptonica PROPERTIES VERSION   ${VERSION_PLAIN})
 set_target_properties           (leptonica PROPERTIES SOVERSION 5.1.0)


DanBloomberg:
The first issue was explored in #253:
   https://github.com/danboomberg/leptonica/issues/253

and much as I'd prefer if we have a single name for the library, it was decided for practical reasons not to change either of the two sonames.  As long as you make both tesseract and leptonica with either autotools or cmake, there should be no issue.

@egorpugin 
What about the optimization level in cmake?

egorpugin:
`-O2` by default in release.

Sineaggi:
Re the naming, the points in #253 are valid. We compile from source so this issue does not affect us. I was under the impression that the cmake install should look as similar to autotools output as possible.

Sineaggi:
The `-O3` to `-O2` switch was made because leptonica fails to build with cmake and `-DCMAKE_BUILD_TYPE=Release` on alpine linux. This also brings us closer to how autotools compiles.

egorpugin:
> I was under the impression that the cmake install should look as similar to autotools output as possible.

This was not a goal.
Goal is to simplify any process, any human actions etc. - always.

Sineaggi:
@egorpugin Sounds good. I've rebased and removed the library renaming commit.

Sineaggi:
If we had the line `set_target_properties (leptonica PROPERTIES OUTPUT_NAME leptonica)` we could produce a pkg-config file on unix platforms without the `Libs: -L${libdir} -lleptonica_OUTPUT_NAME-NOTFOUND` mentioned in #253

xsacha:
Doesn't the default CMake release flag contain -O2 anyway?
What sets it to O3?

DanBloomberg:
Egor said it is -O2 in #279

egorpugin:
I'm going to merge this, as I've checked some build and the O level is really -O3, not -O2.

PR #282, opened 2017-10-03
Set {,DEBUG_}OUTPUT_NAME on non-Win32 platforms

cysp:
… so that lept.pc can be created with the correct linker information.


egorpugin:
We had some discussion on this, so that *nix version library must be 'lept' (or processed as 'liblept'). But this is true for official autotools build and not true for cmake build.
So, I'm going to merge cmake changes on behalf of cmake users' requests.

PR #283, opened 2017-10-24
add support for openjpeg 2.3

BorisMansencal:
Small changes to add support for openjpeg 2.3.
This is in particular necessary to compile on msys2.


DanBloomberg:
@egorpugin 

Boris, thank you for this.  From what I've seen, they have made significant improvements in the encoding efficiency in openjpeg 2.3.

 I'll let Egor sign off on the cmake diff.

PR #286, opened 2017-11-23
Minor memory leaks, performance issues and logic abreviation made.

Carmot:
Also changed pi values across the code to M_PI constant defined in math.h

### DIFF
--- a/prog/affine_reg.c
+++ b/prog/affine_reg.c
@@ -42,7 +42,7 @@ static l_int32 RenderHashedBoxa(PIX *pixt, BOXA *boxa, l_int32 i);
      *    5: test with large distortion
      */
 static const l_int32  x1[] =  { 300,  300,  300,  95,    32};
-static const l_int32  y1[] =  {1200, 1200, 1250, 2821,  934};
+static const l_int32  yy1[] =  {1200, 1200, 1250, 2821,  934};
 static const l_int32  x2[] =  {1200, 1200, 1125, 1432,  487};
 static const l_int32  y2[] =  {1100, 1100, 1100, 2682,  934};
 static const l_int32  x3[] =  { 200,  200,  200,  232,   32};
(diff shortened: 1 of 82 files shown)


DanBloomberg:
Thanks for this.  I'm sure most of these changes are good.  What tool did you use to derive this?

I have not yet gone over this in detail.  I have the following requests:
(1) break it up into 2 parts: changes to the library and changes to the programs
(2) look over it in detail to see where the static analysis failed, and remove them.  For example, the change in list.c, treating pointers as if they were booleans, would be disastrous.
(3) also separate out the changes where it is suggesting a change (e.g., on an index) but not actually making the change -- so actually, I'd like you to break this change into 3 parts.

Then I can deal with the library first, then the suggested changes that are comments, and finally the programs.

  -- Dan

DanBloomberg:
line 1728 of grayquant.c is another example of a disastrous substitution.

DanBloomberg:
on second thought, since i have to go over this in detail anyway, don't bother with making the changes i suggested.  However, if you want to look it over and note changes (like the ones mentioned above) that are incorrect, that would be very helplful.

my general approach to bulk changes like these is to do them incrementally in the library, running the full set of regression tests periodically as the changes are put in.

DanBloomberg:
I've looked at about half the suggested changes for the library, and have committed some of the changes. The conflicting files above are due to that.

Aside from the logic errors with pointers that the static checker made, some other issues that came up are:
  * Can't use M_PI (double) generally; pta functions use floats.
  * graylevels checked in pixOctcubeQuantMixedWithGray(); must be > 1.
  * in pageseg.c, need all "if (pixadb) { " because of local variable declarations

Many of the float casts are unnecessary; the ones I kept involved preventing integer division and avoiding overflow with integer multiplication.  In all, I've kept about half of the suggested changes.

I expect to finish going over the library issues within a week or two.

PR #293, opened 2018-01-14
Use portable fstatat instead of dirent.d_type

ip1981:
Portability means source code portability (this is true that Solaris-based systems lack `d_type`) and runtime portability which means that the application should always test for `d_type == DT_UNKNOWN` (some filesystems may not set `d_type`).

Ref. https://www.osdyson.org/projects/dyson/wiki/Particular_recipes#Missing-d_type-of-struct-dirent

### DIFF
--- a/src/sarray1.c
+++ b/src/sarray1.c
@@ -132,6 +132,7 @@
 #include <string.h>
 #ifndef _WIN32
 #include <dirent.h>     /* unix only */
+#include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


ip1981:
Other option is to delete this code because:

1) it does not compile sometimes
2) if compiled, sometimes does not work as expected

:)

DanBloomberg:
Sorry I didn't get around to this sooner.   This is very nice, and much appreciated.

  -- Dan

PR #296, opened 2018-01-29
Add boxa1_reg to gnuplot-dependent test list

chewi:


DanBloomberg:
I haven't been keeping that file up to date as I've been adding reg tests to alltests_reg.c.
Should they be kept in sync?
If so, there may be several in alltests_reg.c that need to be added.

chewi:
This is just the list of tests that use gnuplot. It is now correct as of v1.75.0.

PR #300, opened 2018-02-01
Fix some compiler warnings

stweil:


DanBloomberg:
Thanks.  I don't think these Wsign warnings are important.  I much prefer using signed ints wherever possible, because with unsigned ints you can get nasty underflows (for example, with -1).  So you have to be much more careful when using signed ints.  Here, I'm mostly using int32 for numbers between 0 and 255, so no possibility of overflow  :-)

In adaptmap, I should probably have just made val and lastval int32.  that would remove the warning.
And I find the warning in blend.c just irritating.

So I think we should remove blend.c and adjust adaptmap so that those two uint32 variables are int32.
Do you want to do that, or should I?

stweil:
I removed the two controversial commits for now. Generally, I personally prefer using unsigned integers for loop counters which often start at 0 or 1 and count up, for indexing of arrays, for other values which are naturally non-negative (width, height, number of elements, ...) and wherever some standard API requires that (most often size_t arguments or return values). Some people also prefer unsigned because it can be faster when divided by a constant. But I am aware that there are different opinions on that question.

DanBloomberg:
thanks!  I'm afraid I have a deep habit of signed loop counters (e,g., over an array of pixels) ...

PR #304, opened 2018-02-08
giflib version check and other configure improvements

chewi:


DanBloomberg:
Really nice job!!!

DanBloomberg:
This tests out perfectly.

there is one puzzle.  config_auto.h has an extra entry for the openjpeg library, which is commented out.  The line for LIBJP2K is the only one that is meaningful.  The LIBOPENJP2 token isn't referenced in the library, so it doesn't do any harm, but I don't know where it came from.

I couldn't attach the file, but here are the relevant lines:


/* Define to 1 if you have libopenjp2. */
#define HAVE_LIBJP2K 1

...

/* Define to 1 if you have the `openjp2' library (-lopenjp2). */
/* #undef HAVE_LIBOPENJP2 */

DanBloomberg:
I can guess that it was manufactured by automake based on the name of the library, which is libopenjp2.

In any event, it does no harm, so if it would be hard to get rid of, don't bother.

chewi:
Fixed that in #306.

PR #306, opened 2018-02-09
Set some configure.ac macro actions to noop where appropriate

chewi:
The AC_CHECK_LIB default action-if-found was defining HAVE_LIBOPENJP2, which isn't used.

I did try this before but got a syntax error. I guess there was something else wrong because it works now.


DanBloomberg:
Thank you again!!  I'll add italic_reg and wordboxes_reg to the gplot list in the wrapper

Issue #307, opened 2018-02-11
webpio_reg test failure on big endian (ppc64, s390x)


manisandro:
Hi,
Noticed this when attempting to update the Fedora package to 1.75.2 (https://koji.fedoraproject.org/koji/taskinfo?taskID=24954012)

    FAIL: webpio_reg
    ================
    ################   webpio_reg   ###############
    leptonica-1.75.2
    libgif 5.1.4 : libjpeg 6b (libjpeg-turbo 1.5.3) : libpng 1.6.34 : libtiff 4.0.9 : zlib 1.2.11 : libwebp 0.6.1
    Time to read jpg:   0.000
    Time to write webp:   0.005
    Time to read webp:   0.000
    Time to read jpg:   0.000
    Time to write webp:   0.000
    Time to read webp:   0.000
    Time to read jpg:   0.000
    Time to write webp:   0.069
    Time to read webp:   0.010
    Time to read jpg:   0.009
    Time to write webp:   0.515
    Time to read webp:   0.049
    qual = 50, psnr =  43.501
    qual = 75, psnr =  46.071
    qual = 90, psnr =  52.379
    qual = 100, psnr =  57.460
    Lossless write:   0.871 sec
    lossless; psnr should be 1000: psnr =  23.222
    Failure in webpio_reg: value comparison for index 12
    difference = 976.777893 but allowed delta = 0.100000
    Time:   3.986 sec
    ################################################

DanBloomberg:
I don't understand the variability.  Here's what I get on linux, also using 0.6.1

leptonica-1.75.2
libgif 5.1.1 : libjpeg 8 : libpng 1.2.35 : libtiff 4.0.3 : zlib 1.2.8 : libwebp 0.6.1 : libopenjp2 2.3.0

Time to read jpg:   0.000
Time to write webp:   0.000
Time to read webp:   0.000
Time to read jpg:   0.000
Time to write webp:   0.004
Time to read webp:   0.000
Time to read jpg:   0.000
Time to write webp:   0.032
Time to read webp:   0.000
Time to read jpg:   0.020
Time to write webp:   0.160
Time to read webp:   0.020
qual = 50, psnr =  43.203
qual = 75, psnr =  45.782
qual = 90, psnr =  51.077
qual = 100, psnr =  54.635
Lossless write:   0.468 sec
lossless; psnr should be 1000: psnr = 1000.000
Time:   1.476 sec

[...]

PR #308, opened 2018-02-13
Skip gnuplot tests if it is missing PNG support

chewi:
I was getting failures when gnuplot was built without it.


DanBloomberg:
Thank you James!  I expect to release patch 3 today.

egorpugin:
Not too fast? :)
Later it can become one release per commit.

chewi:
No rush here, I've already patched it on Gentoo.

DanBloomberg:
:-)

Jeff needs the final debian version out this week, for inclusion in the major ubuntu 18.04 release.

Patch 3 should fix a serious bug with webp support on big-endian machines.  If I don't screw up the tarball, we should be done with 1.75.

DanBloomberg:
The patch was no good, so I've essentially reverted it.

We still have an outstanding issue (#307 ) with lossless webp encoding on big-endians.

PR #310, opened 2018-02-17
README.md: Fix link

amitdo:
jbig2enc: The (previous) link redirects to another url, so use that url directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ www.leptonica.org
 ## Open Source Projects that use Leptonica ##
   * [tesseract](https://github.com/tesseract-ocr/tesseract/) (optical character recognition)
   * [OpenCV](https://github.com/opencv/opencv) (computer vision library)
-  * [jbig2enc](http://www.imperialviolet.org/jbig2.html) (encodes multipage binary image documents with jbig2 compression)
+  * [jbig2enc](https://github.com/agl/jbig2enc) (encodes multipage binary image documents with jbig2 compression)
   * [php](http://en.wikipedia.org/wiki/PHP)  (scripting language for dynamic web pages)
 
 ## Major contributors to Leptonica ##


amitdo:
Does PHP standard library really uses Leptonica?

DanBloomberg:
@amitdo 
As of about 5 years ago, php was using leptonica for image i/o.
I do not know if that is still the case.
If not, we should remove that reference.

amitdo:
Dan, what about the PR?

DanBloomberg:
I was waiting for it to go green.  :-)

amitdo:
about php:

>As of about 5 years ago, php was using leptonica for image i/o.

I didn't find anything in their current tree or 4.x/5.x/7.x Changelog that refers to leptonica.

It use libgd for image i/o

https://github.com/php/php-src/tree/PHP-7.2.0/ext/gd

https://en.wikipedia.org/wiki/GD_Graphics_Library

>GD is extensively used with PHP, where a modified version supporting additional features is included by default as of PHP 4.3 and was an option before that. As of PHP 5.3, a system version of GD may be used as well, to get the additional features that were previously available only to the bundled version of GD

amitdo:
http://php.net/ChangeLog-7.php
http://php.net/ChangeLog-5.php
http://php.net/ChangeLog-4.php

DanBloomberg:
I went back to my correspondence on this with Antony Dovgal, who was
maintaining php in 2007/2008, and he was using leptonica and submitting bug
reports & fixes.

So it was 10 years ago, not 5  (how time flies!).  We should remove that
line from README.md.

amitdo:
https://github.com/tony2001/leptonica

DanBloomberg:
It was just an application.  For some reason I thought Tony was a php library maintainer.

amitdo:
>For some reason I thought Tony was a php library maintainer.

https://pecl.php.net/user/tony2001
PECL - the PHP Extension Community Library.
Similarly, python has PyPI -  the Python Package Index.
https://pypi.python.org/.
Both are a collection of 3rd party packages hosted in the main (PHP/Python) site.

Maybe that was the source of the confusion.

amitdo:
Here are two bindings for leptonica:

https://github.com/cosmo0920/leptonica-capi-go (Apache 2.0)
https://github.com/tvn-cosine/leptonica.net/tree/master/leptonica.net (GPL)

Issue #303, opened 2018-02-06
gplotMakeOutput Command Injection Vulnerability


kbabiochSUSE:
An exploitable command injection vulnerability exists in the gplotMakeOutputfunction of Leptonica 1.74.4. A specially crafted gplot rootname argument can cause a command injection resulting in arbitrary code execution. An attacker can provide a malicious path as input to an application that passes attacker data to this function to trigger this vulnerability.

For details refer to [this][1] write-up. This has been assigned CVE-2018-0516.

[1]: https://www.talosintelligence.com/vulnerability_reports/TALOS-2018-0516

DanBloomberg:
That command injection vulnerability has been patched in 1.75.1.

kbabiochSUSE:
The commit in question fixed several issues: b88c821f8d347bce0aea86d606c710303919f3d2

For those interested in a backport, have a look at [this](https://build.opensuse.org/package/view_file/home:kbabioch:branches:openSUSE:Leap:42.3:Update/leptonica/CVE-2018-3836.patch?rev=50c761a0edab801f01257d5d4f7b04c6).

regiwils:
This is the correct CVE-2018-3836 for this issue

bwhacks:
This doesn't seem to be fixed, as the "bad character" check still allows e.g. "$(command)".

carnil:
FTR, the issue as raised by @bwhacks in https://github.com/DanBloomberg/leptonica/issues/303#issuecomment-366472212 got assigned CVE-2018-7440

DanBloomberg:
@carnil 

And it was fixed 4 hours ago on the master.
Is there anything further that needs to be done?

santiagorr:
Also related to this issue: gplotMakeOutput function does not block either the '/' character.
CVE-2018-7442 has been assigned for a potential path traversal and arbitrary file overwrite.

DanBloomberg:
I don't understand this comment.

The '/' character is required for paths.  The fix in gplotCreate() guarantees that gplot->cmdname is a legitimate path.

egorpugin:
As I understand it will be much safer to not use `system()` calls.
It's very very hard to protect and prepare command line for it.
Better to call via `exec*` family or other OS specific calls.

Am I missing something?

DanBloomberg:
I admit to not knowing anything about the differences in security when using system as opposed to fork/exec/wait.

Can someone point me to some high-level references on this issue?

[...]

PR #313, opened 2018-02-23
Check for $() in gplotCreate()

kbabioch:
This adds $() to the list of bad characters to look for in src/gplot.c to
prevent `$(command)` from being executed as reported in #303.

However, this should not be considered a real fix, since it violates rule 2 of
the [Six Dumbest Ideas in Computer Security][1], namely "Enumerating badness".

[1]: http://www.ranum.com/security/computer_security/editorials/dumb/

### DIFF
--- a/src/gplot.c
+++ b/src/gplot.c
@@ -153,7 +153,7 @@ GPLOT   *gplot;
     if (outformat != GPLOT_PNG && outformat != GPLOT_PS &&
         outformat != GPLOT_EPS && outformat != GPLOT_LATEX)
         return (GPLOT *)ERROR_PTR("outformat invalid", procName, NULL);
-    stringCheckForChars(rootname, "`;&|><\"?*", &badchar);
+    stringCheckForChars(rootname, "`;&|><\"?*$()", &badchar);
     if (badchar)  /* danger of command injection */
         return (GPLOT *)ERROR_PTR("invalid rootname", procName, NULL);
 


DanBloomberg:
Hi Karol,

Thanks for the patch and the link.  It's actually not so dumb -- it's called "engineering".

kbabioch:
Don't want to be too philosophical about it, but it actually is dumb. Because we are fixing the same vulnerability over and over again, each time adding some characters, after some one has another clever idea on how to execute commands. There might always be another clever hack with some kind of escaping, encoding, etc.

It would be much more appropriate to whitelist commands that are actually allowed at this point. Unfortunately I don't know anything about leptonica and how it interacts with the system, but letting users invoke random commands always asks for trouble.

And to be quite honest this has nothing to do with engineering. Engineering is about applying best practices to problems. We are ignoring best practices here!

DanBloomberg:
I agree that engineering is about applying best practices, but best practices doesn't necessarily mean best security.  All of engineering is a tradeoff between cost and benefit.

What would you consider to be 'best practices' for this function?

kbabiochSUSE:
> but best practices doesn't necessarily mean best security.

That's certainly true, but in this case (as with all of the other CVEs filed against leptonica), its mostly ignorance.

> What would you consider to be 'best practices' for this function?

As previously said: A whitelist of commands that can/should/must be run at this point. Only allowing commands that are actually needed is much more secure than trying to prevent bad commands from being run.

DanBloomberg:
I'd be grateful for a patch that would follow modern security best practices on linux and windows.

DanBloomberg:
We've made a number of changes to solve the security issues discussed in this thread.

The leptonica library can now in principle be shipped in a configuration where it is impossible to:
(1) make a subdirectory of /tmp
(2) write anything to a named file known to the compiler, such as "/tmp/lept/foo.png".
(3) execute any program (e.g., display or gnuplot) from the library.
The sole exception is that regression testing does allow this.  However, nobody will run regression testing in production (e.g., in tesseract).

We have also allowed a programmer to write a testing or debug program that does enable writing to a named file in /tmp or a subdirectory of /tmp, or running gnuplot or an image display program.  But again, that wouldn't be done for any program in a production environment, where there is never a requirement for any of those debug and testing operations.

This is all in the master at head.  We have not yet made an official release (1.76) with these features, but I want to bring these security changes to your attention.

kbabioch:
Do you have an plan / ETA for next release?

DanBloomberg:
Hi Karol,

I can do it within the next 2 weeks.  I'm on vacation this week.

  -- Dan

DanBloomberg:
Just released 1.76.0.

DanBloomberg:
I believe the issues are resolved.  See comments at the top, which are actually valid for 1.76.0 as well:

     https://github.com/DanBloomberg/leptonica/blob/master/version-notes.html

PR #320, opened 2018-03-03
Fix CID 1365421, CID 1365459 (Uninitialized scalar variable)

stweil:


DanBloomberg:
Thank you.

PR #324, opened 2018-03-10
Fix some compiler warnings for MS C

stweil:
The MS compiler complains each time when pix.h is included.
When building tesseract, this contributes more than 300 warnings:

src\pix.h(243): warning C4305: 'initializing':
 truncation from 'double' to 'l_float32' (compiling source file ...
src\pix.h(245): warning C4305: 'initializing':
 truncation from 'double' to 'l_float32' (compiling source file ...

The compiler does not complain for line 244, maybe because
a 32 bit float can store 0.5 without any truncation.

Using float constant values should fix those warnings.

### DIFF
--- a/src/pix.h
+++ b/src/pix.h
@@ -240,9 +240,9 @@ enum {
  *          International Telecommunications Union, via ITU-R.
  * </pre>
  */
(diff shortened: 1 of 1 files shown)


egorpugin:
Merging myself because it's trivial.

DanBloomberg:
Thanks, both of you  :-)

Issue #327, opened 2018-03-18
fstatat() not available on all platforms.


Ionic:
In 2b961263329463d29745259f1432f2197f294666, a non-portable variant of checking for directories has been replaced with calls to `fstatat()`.

While that is generally favorable, `fstatat()` is not available on all platforms, examples include pre-10.10 macOS systems.

PR #328, opened 2018-03-18
Add `fstatat()` availibility autodetection/configuration and replace with `stat()` if necessary

Ionic:
Fixes: DanBloomberg/leptonica#327

### DIFF
--- a/src/sarray1.c
+++ b/src/sarray1.c
@@ -1855,7 +1855,7 @@ l_int32         len;
 SARRAY         *safiles;
 DIR            *pdir;
 struct dirent  *pdirentry;
-int             dfd;
+int             dfd, stat_ret;
 struct stat     st;
 
     PROCNAME("getFilenamesInDirectory");
@@ -1871,9 +1871,14 @@ struct stat     st;
     safiles = sarrayCreate(0);
     dfd = dirfd(pdir);
     while ((pdirentry = readdir(pdir))) {
+#if HAVE_FSTATAT
+      stat_ret = fstatat(dfd, pdirentry->d_name, &st, 0);
+#else
+      stat_ret = stat(pdirentry->d_name, &st);
+#endif
 
         /* It's nice to ignore directories.  */
-      if ((0 == fstatat(dfd, pdirentry->d_name, &st, 0))
+      if ((0 == stat_ret)
           && S_ISDIR(st.st_mode)) {
             continue;
       }
--- a/README.html
+++ b/README.html
@@ -500,7 +500,18 @@ that you are using the static makefiles and customizing environ.h.
    If you're building with the autoconf programs, these two functions are
    automatically enabled if available.
 
-3. Typedefs
+3. Runtime functions not available on all platforms
+
+   Some functions are not available on all systems.  One example of such a
+   function is fstatat().  If possible, such functions will be replaced by
+   wrappers, stubs or behavioral equivalent functions.  By default, such
+   functions are disabled; enable them by setting #define HAVE_FUNC  1 (in
+   environ.h).
+
+   If you're building with the autoconf or cmake programs, these functions are
+   automatically enabled if available.
+
+4. Typedefs
 
    A deficiency of C is that no standard has been universally
    adopted for typedefs of the built-in types.  As a result,
@@ -543,7 +554,7 @@ that you are using the static makefiles and customizing environ.h.
    necessary we use the typedefs in stdint.h to specify the pointer
    size (either 4 or 8 byte).
 
-4. Compile-time control over stderr output (see environ.h)
+5. Compile-time control over stderr output (see environ.h)
 
    Leptonica provides both compile-time and run-time control over
    messages and debug output (thanks to Dave Bryan).  Both compile-time
@@ -554,7 +565,7 @@ that you are using the static makefiles and customizing environ.h.
    NO_CONSOLE_IO is defined on the compile line.  For production code
    where no output is to go to stderr, compile with -DNO_CONSOLE_IO.
 
-5. In-memory raster format (Pix)
+6. In-memory raster format (Pix)
 
    Unlike many other open source packages, Leptonica uses packed
    data for images with all bit/pixel (bpp) depths, allowing us
@@ -575,7 +586,7 @@ that you are using the static makefiles and customizing environ.h.
    and a DPix is provided for 2D arrays of doubles.  Converters
    between these and the Pix are given.
 
-6. Conversion between Pix and other in-memory raster formats
+7. Conversion between Pix and other in-memory raster formats
 
  . If you use Leptonica with other imaging libraries, you will need
    functions to convert between the Pix and other image data
@@ -584,7 +595,7 @@ that you are using the static makefiles and customizing environ.h.
    ordering and byte ordering on raster lines.  See the file pix.h
    for the specification of image data in the pix.
 
-7. Custom memory management
+8. Custom memory management
 
    Leptonica allows you to use custom memory management (allocator,
    deallocator).  For Pix, which tend to be large, the alloc/dealloc
--- a/cmake/Configure.cmake
+++ b/cmake/Configure.cmake
@@ -73,6 +73,7 @@ check_includes(include_files_list)
 
 set(functions_list
     fmemopen
+    fstatat
 )
 check_functions(functions_list)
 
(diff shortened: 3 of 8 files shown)


DanBloomberg:
Thank you for fixing the problem introduced in January with fstatat().

Ionic:
You're welcome.

There is one caveat, though: if you were to rely on `fstatat()`-specific behavior (like passing non-zero flags or relying on it treating a relative path relative to the file descriptor's current location instead of current to cwd), then `stat()` is not a proper replacement for `fstatat()`.

To the best of my knowledge, as far as looking through your code went, I couldn't determine anything relying on `fstatat()`-specific behavior, but if the code is changed developers should definitely keep that in mind.

src/sarray1.c
```diff
@@ -1871,9 +1871,14 @@ struct stat     st;
     safiles = sarrayCreate(0);
     dfd = dirfd(pdir);
     while ((pdirentry = readdir(pdir))) {
+#if HAVE_FSTATAT
+      stat_ret = fstatat(dfd, pdirentry->d_name, &st, 0);
+#else
+      stat_ret = stat(pdirentry->d_name, &st);
```
ip1981: This depends on the current directory.

ip1981:
I could not care less about old non-free OSes, but if this is a problem, removing this code is the next best option :) Either unconditionally or when `fstatat` is absent.

DanBloomberg:
OK, help me out here.  When you say "removing this code is the next best option", what code are you referring to?

 I really have 3 questions:
* does anything need to be changed, and if so, why?
* do we have consensus on that?

  -- Dan

ip1981:
I refer to the code that detects directories. How useful is it?

Existing code should work just fine with `fstatat`, but not with `stat` .
This PR addresses compilation issues only.

I remember I fixed both compilation and runtime issues (reiserfs AFAIK does not set  `d_type` on Linux, no `d_type` on Solaris).

So again, how critical is this feature (skipping on directories)?

DanBloomberg:
It may not be very important for most applications.  But if the information is available, why not use it?

I did some experiments.  As you say, it all works with fstatat.
But with stat, it finds "." and ".." as directories but not subdirectories.

Why not just put in a comment to that effect?

DanBloomberg:
I did that & pushed to the master.   Are we ready to close this issue?

Ionic:
`stat()` and `fstatat()` behave exactly the same if passed an absolute path (assuming the `flags` are `0`). Using `realpath` on `realdir` and prepending this to additional directory entries when calling `stat()` (only) would work around any dependence upon `CWD`.

DanBloomberg:
When I do this (calling stat, not fstatat):

    char *realp = realpath("lion", NULL);
    Sarray *sa = getFilenamesInDirectory(realp);

I still get the subdirectories of "lion" added into sa.
What flags are supposed to be 0?

Ionic:
Careful with passing `NULL` to `realpath()`. This invokes implementation-defined behavior [in older versions of SUS/POSIX](http://pubs.opengroup.org/onlinepubs/009695399/functions/realpath.html) and some implementions crash in such a case. [Newer versions allow this, making `realpath()` malloc a buffer by itself](http://pubs.opengroup.org/onlinepubs/9699919799/functions/realpath.html), but of course that is not widely supported yet. It's better to always give it a proper `PATH_MAX + 1`-length buffer. (That, btw., is in line with the lot of `str*` functions that expect proper strings and often crash if passing `NULL`.)

`When I do this (calling stat, not fstatat):`

You'll also have to prepend the absolute base path to the filename when calling stat, i.e.,

```C
#if HAVE_FSTATAT
        stat_ret = fstatat(dfd, pdirentry->d_name, &st, 0);
#else
        size_t tmp_size = strlen(realdir) + strlen(pdirentry->d_name) + 1; /* Additional slash */
        char *stat_path = malloc(tmp_size + 1); /* Terminating NULL */
        snprintf(stat_path, tmp_size + 1, "%s/%s", realdir, pdirentry->d_name);
        stat_ret = stat(stat_path, &st);
        free(stat_path);
        stat_path = NULL;
#endif
```

DanBloomberg:
Thanks!  That worked well.  I think this resolves the issue, as now stat() and fstatat() both handle directories properly.

Ionic:
Yeah, but `realpath()` should also be executed unconditionally in the `stat()` case, since the function shouldn't rely upon being provided an absolute path... at least in theory.

ip1981:
Oh, no :) multiple allocations and deallocations without checking for NULL. Is the problem really worth it?

Ionic:
That was just meant as a quick way for Dan to demonstrate what I was referring to.

Naturally in the actual code the return values of `malloc()`, `snprintf()` and `realpath()` (which should be inserted into that function in the `stat()`-branch as well, instead of assuming that an actual absolute path is always passed in) should all be checked and the function error'd out properly.

Whether or not that's worth it... I guess it depends. Changing behavior if `fstatat()` is not available sounds worse to me than a compat shim. While `fstatat()` was commonly available on Linux-based platforms, `*BSD` ones only adopted out ~10 years ago. That's a long time span, but why break stuff if it's so easy to provide an alternative?

DanBloomberg:
Thank you again for all the help.  I've now made changes so that it gets the same (correct) results with either stat() or fstatat().

It doesn't need to check for the return of realpath() because we're not allocating memory there -- just writing to a buffer on the stack.

Are we ready to close this issue?

Ionic:
Should be `PATH_MAX + 1` (since the macro does not include a NULL terminator).

Like @ip1981 suggested, it's also worthwhile to
```C
    if (!stat_path)
        return (SARRAY *)ERROR_PTR("unable to allocate memory for stat_path", procName, NULL);
```

after the `LEPT_CALLOC` call for `stat_path`.


Plus, check `snprintf`'s return code:
```C
    size_t snprintf_ret = snprintf(stat_path, size - 1, "%s/%s", realdir, pdirentry->d_name);
    if ((snprintf_ret < 0) || (snprintf_ret >= size)
        return (SARRAY *)ERROR_PTR("unable to construct absolute path", procName, NULL);
```

There is also a problem with the size expression provided there: you add `+ 2` for both the slash and `NULL` terminator. In that case, you can just alloc `size` (not `size + 1` bytes), but will have to pass `size - 1` to `snprintf`, since the `n` denotes the number of characters that are to written minus the always written NULL byte (yeah, `snprintf` is a bit hairy). I've corrected this in my previous code example for the `snprintf` call - the return value check is also correct, because it denotes the bytes that would have been written if `n` was large enough, excluding the terminating NULL byte (hence, if it's returning a number of `size` or bigger you know that the output was truncated).

Otherwise I'm content.

Ionic:
Should have been `ptrdiff_t` instead of `size_t` of course, since the return code can be negative.

DanBloomberg:
Thank you again for reviewing the code. It is finicky to get this done well!

I took some of your advice, and did it a bit differently.
I checked the size against PATH_MAX and skipped with an error message if
too large.
Otherwise, we know the calloc will succeed (it's small), so no need to
check that.
The snprintf() needs size, not size - 1, to get all the characters, and
again no need to check the return since we know the buffer is big enough
(it is PATH_MAX + 1, as you suggested).

Ionic:
> I checked the size against PATH_MAX and skipped with an error message if too large.

Okay, sane.

> Otherwise, we know the calloc will succeed (it's small), so no need to check that.

Well... strictly speaking, memory exhaustion (i.e., big requests) is not the only cause of `malloc` failures; if the program's memory space is heavily fragmented, allocating even smaller requests may fail. I typically check all allocations, no matter how small, but it's your call of course. :)

> The snprintf() needs size, not size - 1, to get all the characters

Right, there was that... it's the total buffer size including the terminating `NULL` character, sorry. I typically check `snprintf` return codes as well, even when I know that the buffer is big enough, but that might be overly defensive.

Anyway, I guess I'm fine with all of that. Unless Igor has other comments, we're probably done here.

ip1981:
There is gnulib which provides drop-in replacement for many functions, including for `fstatat` :)

Anyway, I think this feature is not worth that much of [sloppy] code. As I mentioned in my original PR I'd delete it.

Depending on the use case, there might be another quite portable option - `ftw`, http://pubs.opengroup.org/onlinepubs/007904975/functions/ftw.html

DanBloomberg:
I agree that a lot of time has been spent on this, but it has been educational for me, and the code is now working correctly in all cases.

From the beginning, I thought of leptonica as having two main goals: (1) the didactic one of showing how to make useful image processing and analysis code (with decent C programming), and (2) providing a library that can be used for both development (because of simplicity) and production (because of efficiency).  Getting this function to work properly seems to satisfy both goals, and I appreciate the effort you both made to get it to that point.

PR #330, opened 2018-04-09
Add some missing test programs to list of other programs

stweil:


DanBloomberg:
These may have been (accidentally or on purpose) omitted because of some security concerns.  I'll check with Jeff and get back within a couple of days  (on vacation).

stweil:
Maybe we need some extra list for such programs then because they are built, so they should also be removed when someone runs `make clean` or `make distclean` (that's how I found them).

DanBloomberg:
Thanks, Stefan.  These were removed 2 months ago because of security concerns over the 'system' call.

I will have these programs output some text when they run, warning that they should only be used in a testing environment because of the possible vulnerability.

DanBloomberg:
So this issue is fixed and I think is OK from a security viewpoint.

PR #332, opened 2018-04-11
fix broken include paths on msvc

dschoerk:
LeptonicaConfig.cmake include directories are formatted wrongly

The current cmake files result in semicolon separated paths:
`set(Leptonica_INCLUDE_DIRS C:/ProgramFiles/leptonica/include;C:/ProgramFiles/leptonica/include/leptonica)`
I am unsure where the problem is exactly here, but as you can see the spaces of "Program Files" had been removed, and the paths are not quoted correctly. This PR produces now include paths formatted as below.

`set(Leptonica_INCLUDE_DIRS "C:/Program Files/leptonica/include" "C:/Program Files/leptonica/include/leptonica")`

i was building with Visual Studio 2015 and CMake 3.11

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,7 +108,8 @@ include(Configure)
 
 configure_file(${AUTOCONFIG_SRC} ${AUTOCONFIG} @ONLY)
 
-set(INCLUDE_DIR "${CMAKE_INSTALL_PREFIX}/include" "${CMAKE_INSTALL_PREFIX}/include/leptonica")
+set(INCLUDE_DIR_INC "${CMAKE_INSTALL_PREFIX}/include")
+set(INCLUDE_DIR_INCLEPT "${CMAKE_INSTALL_PREFIX}/include/leptonica")
 
 configure_file(
     ${CMAKE_SOURCE_DIR}/cmake/templates/LeptonicaConfig-version.cmake.in
--- a/cmake/templates/LeptonicaConfig.cmake.in
+++ b/cmake/templates/LeptonicaConfig.cmake.in
@@ -36,7 +36,7 @@ SET(Leptonica_VERSION_PATCH     @VERSION_PATCH@)
 # ======================================================
 
 # Provide the include directories to the caller
-set(Leptonica_INCLUDE_DIRS @INCLUDE_DIR@)
+set(Leptonica_INCLUDE_DIRS "@INCLUDE_DIR_INC@" "@INCLUDE_DIR_INCLEPT@")
 
 # ====================================================================
 # Link libraries:


egorpugin:
But those paths do not participate in build.
Am I wrong?

egorpugin:
For me it is:
```
# Provide the include directories to the caller
set(Leptonica_INCLUDE_DIRS C:/Program Files (x86)/leptonica/include;C:/Program Files (x86)/leptonica/include/leptonica)
```
So, I cannot reproduce your output. Can you provide detailed steps?

dschoerk:
you are correct, the named paths do not affect the leptonica build, but the generated cmake configuration affects projects that try to link leptonica. 

i have analyzed the issue now more closely, and the problems i found are caused by spaces contained in the installation path. excuse my partly incorrect initial post. the reason for the omitted spaces there was just caused by using message(...) in cmake. 



include dirs for a leptonica installation path without spaces; (from LeptonicaConfig.cmake: `set(Leptonica_INCLUDE_DIRS ../leptonica-install/include;../leptonica-install/include/leptonica)`)

```
dir='E:/Programming/nomacs-plugins/OcrPlugin/ext/tesseract/../leptonica-install/include'
dir='E:/Programming/nomacs-plugins/OcrPlugin/ext/tesseract/../leptonica-install/include/leptonica'
```

include directories with spaces; (from LeptonicaConfig.cmake: `set(Leptonica_INCLUDE_DIRS ../leptonica install/include;../leptonica install/include/leptonica)`)

```
dir='E:/Programming/nomacs-plugins/OcrPlugin/ext/tesseract/../leptonica'
dir='E:/Programming/nomacs-plugins/OcrPlugin/ext/tesseract/install/include'
dir='E:/Programming/nomacs-plugins/OcrPlugin/ext/tesseract/../leptonica'
dir='E:/Programming/nomacs-plugins/OcrPlugin/ext/tesseract/install/include/leptonica'
```

these directories have been listed with
```get_property(dirs DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES)
foreach(dir ${dirs})
  message(STATUS "dir='${dir}'")
endforeach()```

as you can see from the above example, the include paths are split incorrectly if a space is in the path. that is what this PR fixes. the problem becomes visible on windows especially because the default installation path is "C:/Program Files/...". However, it is more of a design flaw inside cmake than in leptonica, but should still be adressed.

**Comment by egorpugin** [comment]
Ok, I've added quotes to Leptonica_INCLUDE_DIRS.
Now I see:
```
>cmake -P LeptonicaConfig.cmake
-- dir='C:/Program Files (x86)/leptonica/include'
-- dir='C:/Program Files (x86)/leptonica/include/leptonica'
```

Could you please check if  your issue is still present?

dschoerk:
that was my first attempt to solve it too. with an earlier cmake version that seemed to fail (i dont know what the reason was exactly, but thats the only reason i can imagine currently).

anyways, with cmake 3.11 your change seems to work fine and is much simpler 👍

Issue #334, opened 2018-04-23
 undefined macros and syntax error with configure.ac


DanBloomberg:
@chewi 

I've installed the latest version of automake (etc).  Now, when running my autobuild file, I am getting errors:

libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'config'.
libtoolize: copying file 'config/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
libtoolize: copying file 'm4/libtool.m4'
libtoolize: copying file 'm4/ltoptions.m4'
libtoolize: copying file 'm4/ltsugar.m4'
libtoolize: copying file 'm4/ltversion.m4'
libtoolize: copying file 'm4/lt~obsolete.m4'
configure.ac:38: error: possibly undefined macro: AS_IF
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:44: error: possibly undefined macro: AC_DEFINE
configure.ac:46: error: possibly undefined macro: AC_CHECK_LIB
configure.ac:47: error: possibly undefined macro: AC_CHECK_HEADER
configure.ac:50: error: possibly undefined macro: AC_SUBST
configure.ac:52: error: possibly undefined macro: AC_MSG_ERROR
configure.ac:178: error: possibly undefined macro: AC_CHECK_HEADERS
configure.ac:179: error: possibly undefined macro: AC_DEFINE_UNQUOTED
configure.ac:8: installing 'config/compile'
configure.ac:10: installing 'config/missing'
prog/Makefile.am: installing 'config/depcomp'

Has something been deprecated?

I could suppress the error messages using m4_pattern_allow([AS_IF])  etc., but could not fix the problem, because when running ./configure, I get this error, and no Makefile is made:

./configure: line 13070: syntax error near unexpected token `ZLIB,'
./configure: line 13070: `  PKG_CHECK_MODULES(ZLIB, zlib,'

chewi:
Did get your earlier message, will look tonight.

DanBloomberg:
Thanks.  Here's some more info.

Running each command in autobuild, the error messages happen in autoconf, which returns failure (1).

Here are the versions:
> dpkg -l autoconf
autoconf       2.69-11  
> dpkg -l libtool
libtool        2.4.6-2

And running ./configure dies with the syntax error on ZLIB.

Jeff has the same versions of these tools, but he can build from head without errors.

[...]

PR #335, opened 2018-04-24
autobuild: Add check for missing pkg-config

stweil:
Tell users if pkg-config is missing instead of generating an invalid
configure script which fails with strange error messages.

### DIFF
--- a/autobuild
+++ b/autobuild
@@ -6,5 +6,10 @@ libtoolize -c -f || glibtoolize -c -f
 aclocal
 autoheader -f
 autoconf
+if grep -q PKG_CHECK_MODULES configure; then
+  # The generated configure is invalid because pkg-config is unavailable.
+  rm configure
+  echo "Error, missing pkg-config. Check the build requirements."
+  exit 1
+fi
 automake -a -c -f
-


stweil:
This addresses issue #334.

DanBloomberg:
Thanks!  Nice to know we won't have this problem again.

amitdo:
@DanBloomberg

https://github.com/DanBloomberg/leptonica/commits?author=stweil
https://github.com/DanBloomberg/leptonica#major-contributors-to-leptonica

DanBloomberg:
@amitgdo

Good point.  I'll make the addition.

amitdo:
>@amitgdo

It's @amitdo :smile:  (my first name + first two letters of my last name).

... unless my parents gave me a middle name that starts with 'g' and didn't tell me. :laughing:  

>Stefan Weil:**:** Stefan ...

One `:` should be dropped.

stweil:
> One : should be dropped.

Maybe also _GitHub_ instead of _github_ would match their corporate identity.

DanBloomberg:
Done.

@amitdo 

they didn't tell me either -- I fat-fingered the 't' key

PR #336, opened 2018-04-24
ioformats_reg: Fix wrong snprintf and uninitialized variable

stweil:
Commit 374e71e02ee204882921761bba5d2d85dfcb0b4e introduced a bug which
is reported by gcc:

prog/ioformats_reg.c:787:29: warning:
 argument to ‘sizeof’ in ‘snprintf’ call is the same expression as the
 destination; did you mean to provide an explicit length? [-Wsizeof-pointer-memaccess]

Fix also a typo in the 2nd fprintf statement and make sure that there is
always a valid format name:

prog/ioformats_reg.c:774:13: warning:
 ‘tiff_compression_name’ may be used uninitialized in this function [-Wmaybe-uninitialized]

### DIFF
--- a/prog/ioformats_reg.c
+++ b/prog/ioformats_reg.c
@@ -87,7 +87,7 @@ static l_int32 testcomp_mem(PIX *pixs, PIX **ppixt, l_int32 index,
 static l_int32 test_writemem(PIX *pixs, l_int32 format, char *psfile);
 static PIX *make_24_bpp_pix(PIX *pixs);
 static l_int32 get_header_data(const char *filename, l_int32 true_format);
-static void get_tiff_compression_name(char *buf, l_int32 format);
+static const char *get_tiff_compression_name(l_int32 format);
 
 LEPT_DLL extern const char *ImageFileFormatExtensions[];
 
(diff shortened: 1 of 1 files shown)


stweil:
It's a little bit different. A constant string like "text" is allocated by the linker at a fixed location in memory during the whole life time of the process. It is not on the stack.

DanBloomberg:
OK, that makes much more sense.

I noticed that you declared it in 2 places, once at the beginning of get_header_data() and again at the beginning of tiff_compression_name().  Each time with an assignment.  But it would exist in just one place, and is pointing at different compiled string globals.

DanBloomberg:
BTW, in this change you quoted:

prog/ioformats_reg.c:787:29: warning:
argument to ‘sizeof’ in ‘snprintf’ call is the same expression as the
destination; did you mean to provide an explicit length? [-Wsizeof-pointer-memaccess]

I don't see the problem.  sizeof(buf) will return 64, the allocated size.  Isn't that why we use snprintf?

stweil:
You don't get the size of the buffer, but the size of the pointer: `sizeof(char *)` is 4 on 32 bit hosts, 8 an 64 bit hosts.

stweil:
The compiler should complain because `sizeof(buf)` needs `"%zu\n"`, but then it will print `64`, yes.

DanBloomberg:
For printing size_t variables,  there has been a compatibility problem with Windows not supporting %z until recently.  Yes, hard to believe it took until VS2013 before they supported some of these parts of the C99 standard!  So on the rare occasions that I needed to print it, for portability I had to cast to (unsigned long) and then use %lu.
   https://stackoverflow.com/questions/15610053/correct-printf-format-specifier-for-size-t-zu-or-iu

PR #342, opened 2018-05-05
Fix bmpio for sparc64

stweil:
sparc64 (and maybe other RISC architectures, too) require aligned access
to data structures, or the compiler must be told that data is misaligned.

Using a packed struct with GCC fixes the SIGBUS for sparc64.

### DIFF
--- a/src/bmpio.c
+++ b/src/bmpio.c
@@ -51,6 +51,14 @@
 #if  USE_BMPIO   /* defined in environ.h */
 /* --------------------------------------------*/
 
+#if defined(__GNUC__)
+/* BMP_IH is misaligned, therefore we need a packed struct. */
+typedef struct __attribute__((__packed__)) {
+  BMP_FH bmpfh;
+  BMP_IH bmpih;
+} BMP_HEADER;
+#endif
+
     /* Here we're setting the pixel value 0 to white (255) and the
      * value 1 to black (0).  This is the convention for grayscale, but
      * the opposite of the convention for 1 bpp, where 0 is white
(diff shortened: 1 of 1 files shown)


stweil:
Test results for sparc64 with this patch:

    ============================================================================
    Testsuite summary for leptonica 1.76.0
    ============================================================================
    # TOTAL: 112
    # PASS:  96
    # SKIP:  15
    # XFAIL: 0
    # FAIL:  1
    # XPASS: 0
    # ERROR: 0

The test mtiff_reg is still failing.

amitdo:
What's the meaning of XFAIL and XPASS ?

egorpugin:
XFAIL = test is expected to fail.
XPASS = test is expected to fail but passed.

DanBloomberg:
@stweil 
Are lines 161 and 520 of bmpio.c actually used?

stweil:
> [...] problems on 64 bit linux boxes?

The typical Linux boxes use Intel or similar processors. Those processors support misaligned memory access, that means they have no problem reading a 32 bit = 4 byte value from an address which is not divisible by 4. Linux boxes with a Sparc64 processor refuse to do that – unless you tell the compiler that the value is misaligned. That's what my commit does.

>  Are lines 161 and 520 of bmpio.c actually used?

Yes, they are used in combination with line 141 if you build with the GNU compiler or with CLANG. They are not used for builds with MS Visual Studio.

DanBloomberg:
Clever!  The problem occurred because BMP_FH happened to have an odd number of 16 bit ints.

stweil:
Yes, that's the reason for the misaligned 32 bit access which follows those 16 bit values. BMP is a rather old format. I think other file formats take more care to align their data.

PR #352, opened 2018-06-09
 Add "const" attribute to string parameter of sarrayAddString and remove unneeded type casts

stweil:


DanBloomberg:
Did you change the prototype in allheaders.h?

DanBloomberg:
sarrayAddString() can take the parameter L_INSERT, which puts the string directly into the sarray.  You wouldn't do that for a string  that is initialized to a const char value, such as:
    const char  *str = "this is a string";
But you could do it for a string that has been made on the heap, and needs to be free'd.

For such a heap-based string, could we possibly run into problems when later on the array of strings is destroyed?

stweil:
> Did you change the prototype in allheaders.h?

Yes, see [here](https://github.com/DanBloomberg/leptonica/pull/352/files#diff-6afe82e0a13c1756c25f60b60211de8c).

stweil:
> For such a heap-based string, could we possibly run into problems when later on the array of strings is destroyed?

The change here does not change the way how `sarrayAddString` works. It is still possible to add a constant string with `L_INSERT` which will result in an error when the array is destroyed. `L_INSERT` will work only for strings which were allocated on the heap and not free'd anywhere else.

DanBloomberg:
Yes, this is fine.  Thank you for getting rid of those ugly casts.

PR #369, opened 2018-09-11
Fix remaining warnings from lgtm

stweil:


DanBloomberg:
thanks!  Nice not to have lgtm nagging us    :-)

stweil:
I'm sorry, but I only fixed the three warnings. lgtm will still complain about one remaining issue which it considers to be an error, not a warning: the potentially dangerous function `gmtime`.

DanBloomberg:
Even better to have 1 "alert", to remind us that software can never be perfect  :-)

I believe that gmtime can only make an error in a multi-threaded situation.

stweil:
Leptonica can be used in multi-threaded applications, so I decided to fix that one, too.

PR #372, opened 2018-09-18
remove unnecessary calls to lept_mkdir

vidiecan:
## Current Behaviour:
using the newest version 1.76.0 out of the box in our project outputs
```
Info in lept_mkdir: making named temp subdirectory lept/orient is disabled
```

## Expected behaviour

No output.

## Warning

Please note that the fix was not properly tested for all the functions.


DanBloomberg:
Thank you!  Somehow I missed these  :-)

PR #382, opened 2018-10-15
boxfunc5: Fix loop count condition

stweil:
This also fixes a warning from LGTM:

    Comparison is always false because i >= 0.


stweil:
Dan, please review. I only tested building.

stweil:
@DanBloomberg, maybe it would be good to add LGTM to the continuous integration tools to detect such issues early. See https://lgtm.com/projects/g/DanBloomberg/leptonica/ci/.

DanBloomberg:
Thanks.  This is brand new code that hasn't been fully tested, obviously  :-)

PR #383, opened 2018-10-16
Replace insecure gmtime

stweil:
This fixes an error from LGTM:

    Call to gmtime is potentially dangerous


DanBloomberg:
Thank you.  I had no idea how to do the windows version.

DanBloomberg:
@stweil  Got a compiler warning about uninitialized ptr; it's now initialized.

stweil:
That was my fault, sorry. Thank you for fixing it.

PR #384, opened 2018-10-23
Handle temporary directories on iOS

bvirlet:
On iOS, you can't write to /tmp; each application has its own temporary directory. So we rely on the system to tell us where this directory is located.

### DIFF
--- a/src/utils2.c
+++ b/src/utils2.c
@@ -183,6 +183,10 @@
 #include <sys/types.h>
 #endif
 
+#ifdef OS_IOS
+#include <unistd.h>
+#endif
+
 #include <string.h>
 #include <stddef.h>
 #include "allheaders.h"
@@ -2916,8 +2920,10 @@ l_int32  dirlen, namelen, size;
  *      (2) Caller allocates %result, large enough to hold the path,
  *          which is:
  *            /tmp/%subdir       (unix)
- *            [Temp]/%subdir     (windows)
- *          where [Temp] is a path on windows determined by GenTempPath()
+ *            [Temp]/%subdir     (windows, mac, ios)
+ *          where [Temp] is a path determined
+ *             - on windows, mac: by GetTempPath()
+ *             - on ios: by confstr() (see man page)
  *          and %subdir is in general a set of nested subdirectories:
  *            dir1/dir2/.../dirN
  *          which in use would not typically exceed 2 levels.
@@ -2945,7 +2951,22 @@ size_t   pathlen;
         return ERROR_INT("subdir not an actual subdirectory", procName, 1);
 
     memset(result, 0, nbytes);
+
+#ifdef OS_IOS
+    {
+        size_t n = confstr(_CS_DARWIN_USER_TEMP_DIR, result, nbytes);
+        if (n == 0) {
+            L_ERROR("failed to find tmp dir, %s\n", procName, strerror(errno));
+            return 1;
+        } else if (n > nbytes) {
+            return ERROR_INT("result array too small for path\n", procName, 1);
+        }
+        dir = pathJoin(result, subdir);
+    }
+#else
     dir = pathJoin("/tmp", subdir);
+#endif /*  ~ OS_IOS */
+
 #ifndef _WIN32
     path = stringNew(dir);
 #else


DanBloomberg:
I don't know the iOS code base.  Is the test based on n correct? 

What happens if the call fails?  With linux you get an error message and a null return,

src/utils2.c
```diff
@@ -3035,8 +3039,15 @@ char  dirname[240];
 
     PROCNAME("l_makeTempFilename");
 
+ #ifdef __APPLE__
+    size_t n = confstr(_CS_DARWIN_USER_TEMP_DIR, dirname, sizeof(dirname));
+    if ((n <= 0) || (n >= sizeof(dirname))) {
+      strlcpy(dirname, getenv("TMPDIR"), sizeof(dirname));
```
stweil: If this code is executed without `TMPDIR` being set in the environment, it will raise SIGSEGV because a NULL pointer is passed to `strlcpy`.

bvirlet: That's correct. I think on Apple we can remove the fallback to TMPDIR and just use the _CS_DARWIN_USER_TEMP_DIR (http://www.manpagez.com/man/3/confstr/). I have tested on iOS and Mac and it seems to work properly. I'm updating the PR.

DanBloomberg:
@stweil 

Is it now OK to merge this PR?

stweil:
@bvirlet, you said that the problem was on iOS, but your patch checks for `__APPLE__`. Will that change the code for macOS, too? On macOS there is no problem using `/tmp`.

bvirlet:
> @bvirlet, you said that the problem was on iOS, but your patch checks for `__APPLE__`. Will that change the code for macOS, too? On macOS there is no problem using `/tmp`.

According to the documentation I believe the result of confstr() and TMPDIR would be the same on MacOS. We run leptonica with this change on MacOS as well and have not experienced any issue.

I wanted to verify this, though. I wrote the following short program and ran in on MacOS: 

```
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>

int main(int argc, char *argv[], char *envp[]) {
  char buf[1024];
  if (confstr(_CS_DARWIN_USER_TEMP_DIR, buf, 1024) != 0) {
    printf("confstr(_CS_DARWIN_USER_TEMP_DIR) = %s\n", buf);
  }
  printf("TMPDIR                            = %s\n", getenv("TMPDIR"));
}
```

Output:
```
confstr(_CS_DARWIN_USER_TEMP_DIR) = /var/folders/ds/94mr7zn90mq0zb39s7cf6d5c0000gn/T/
TMPDIR                            = /var/folders/ds/94mr7zn90mq0zb39s7cf6d5c0000gn/T/
```

stweil:
The new code for iOS should be in `makeTempDirname`, not in `l_makeTempFilename`.

DanBloomberg:
I haven't forgotten this PR.   Just haven't seen a response to @stweil 's suggestion.

bvirlet:
Hey Dan! Haven't forgotten either, just need to find some time to work on it. I'm not super familiar with the code base so thanks for pointing out the better location for this code. Could you enlighten me on the `l_` prefix? What does it mean? Is it for private APIs?

DanBloomberg:
C doesn't have namespaces, so appending 'l_' is an attempt to avoid issues with other C libraries, even if I think there's only a small chance.

bvirlet:
PR updated. Thanks for the constructive comments!

I left the two independent commits for review, but I will squash them when approved for merge.

DanBloomberg:
Actually, I'd do it a bit differently (with the same result).
C doesn't allow variable declaration except after braces.  So you need to declare
   char  systemTempDir[240];
   size_t  n
at the top,m or use braces.  I'd suggest the latter.  So you'd have something like this (always use 4 space indents):

#ifdef __APPLE__
  {
  char systemTempDir[240]
  size_t n = ...
  if (n == 0 ...)
     return ERROR_INT("falied to find tmp dir", procName, 1);
  dir = pathJoin(systemTempDir, subdir);
  }
#else    
 dir = pathJoin("/tmp", subdir);
#endif  /*  __APPLE __ */

bvirlet:
I have applied your recommendations.

src/utils2.c
```diff
@@ -2945,7 +2951,22 @@ size_t   pathlen;
...
 
     memset(result, 0, nbytes);
+
+#ifdef __APPLE__
+    {
+        size_t n = confstr(_CS_DARWIN_USER_TEMP_DIR, result, nbytes);
+        if (n == 0) {
+            L_ERROR("failed to find tmp dir, %s\n", procName, strerror(errno));
+            return 1;
+        } else if (n > sizeof(nbytes)) {
```
stweil:
```suggestion
         } else if (n > nbytes) {
```

src/utils2.c
```diff
@@ -2945,7 +2951,22 @@ size_t   pathlen;
...
+            return ERROR_INT("tmp directory name too long", procName, 1);
```
stweil: I'd use the same text as in line 2979 here.

bvirlet:
Hi Dan and Stefan! Happy new year! Do you have some feedback on this last change?

src/utils2.c
```diff
@@ -2945,7 +2951,22 @@ size_t   pathlen;
         return ERROR_INT("subdir not an actual subdirectory", procName, 1);
 
     memset(result, 0, nbytes);
+
+#ifdef __APPLE__
```
stweil: It should be possible to restrict that code part to iOS. Other code locations already use `#ifdef OS_IOS`.

stweil:
```suggestion
#ifdef OS_IOS
```

src/utils2.c
```diff
@@ -183,6 +183,10 @@
 #include <sys/types.h>
 #endif
 
+#ifdef __APPLE__
```
stweil:
```suggestion
#ifdef OS_IOS
```

src/utils2.c
```diff
@@ -183,6 +183,10 @@
...
+#include <CoreFoundation/CoreFoundation.h>
```
stweil:
```suggestion
#include <unistd.h>
```

stweil:
@bvirlet, thank you for the updates. I now had some time to review the new code. To summarize: I think it is fine for iOS, but I would not introduce it for macOS. You are right that `TMPDIR` points to the same directory as the result from `confstr`, but `TMPDIR` is not used by Leptonica for macOS.

DanBloomberg:
I'll wait for you to resolve the macOS vs iOS issue (likely restricted to iOS for the new code)

bvirlet:
Before you merge, I'm testing this on iOS and I will squash the commits to keep the history clean.

DanBloomberg:
@stweil, @bvirlet 

where do we stand on this?  Do we believe it works for both iOS and macOS?  should it be restricted to just iOS?

DanBloomberg:
Thank you both for your patience.

bvirlet:
Thank you @DanBloomberg and @stweil for your thorough reviews!

DanBloomberg:
You are welcome.

You might be amused to know that after all that, your change would not compile in some iOS google projects.  It seems that with iOS, error.h needs to be explicitly included in the source file.

PR #387, opened 2018-10-30
Uses CMAKE_CURRENT_ variables intead of CMAKE_ globals

joelgallant:
This allows leptonica to be included as a subdirectory in other projects (my preferred way of handling 3rd party deps). I tested in a local project and everything else does seem to function as intended.


stweil:
s/intead/instead/ in the subject line,

PR #390, opened 2018-11-11
Add const attributes to several function parameters

stweil:
This allows for better compiler optimization and avoids type casts
in code using the Leptonica API.


DanBloomberg:
Hi Stefan,
const is a slippery slope.   I strongly don't want users to ever have to cast it away.
Let me think about this for a few days.

question:   suppose function:    foo(const Pix *p1)
                  foo calls bar(p1), not bar(const p1), but it doesn't change p1
                   does that compile?

                   likewise, suppose Pix contains *cm, where cm is a struct
                  foo calls copycm(cm)       which has the signature
                      copycm(struct *cm)      [not copycm(const struct *cm)]
                  does that work?

stweil:
Neither of both cases would compile. That's why I did not change the prototypes of such functions. Nor does it work passing `const PIX *pix` to functions like `pixGetData` which don't change `pix`, but return parts of it.

Also important: my change is compatible with existing code, so does not require code updates for users of the Leptonica API.

DanBloomberg:
Please help me understand.  All the consts you added seem straightforward, except for  pixResizeImageData(), which calls:
    cmaps = pixGetColormap(pixs)
    pixcmapIsValid(cmaps ...)
    pixcmapCopy(cmaps)
It's true that none of these change cmaps, but none of them declare pixs or its cmap to be const.
In fact,  pixGetColormap() just gives a pointer to the colormap in pixs, so it could never be declared const.

Does the compiler look at these 3 functions and decide that because they are not changing pixs or its cmap, that it's ok, even though they don't declare pixs or its cmap to be const?

DanBloomberg:
So ... is this the "camel's nose under the tent", or is it essentially the complete set of functions for which it makes sense to declare those Pix parameters 'const'?

stweil:
I tried to handle all functions which have a Pix pointer parameter and updated those of them which only needed an added `const` attribute. Some more functions like `picClone` or `pixCopy` also don't change their `pixs` argument, so could also use `const`, but they would currently require type casts internally because they call other functions which don't expect a `const PIX *` argument. Those functions could be fixed without using type casts. That would require using new API functions like `pixGetConstText` in addition to the existing `pixGetText`. Those new functions then must return `const` pointers.

stweil:
Regarding your question about the camel's nose, I first had to look up what it means and found it at [Wikipedia](https://en.wikipedia.org/wiki/Camel%27s_nose). There they speak of "permitting of a small, seemingly innocuous act [which] will open the door for larger, clearly undesirable actions". My pull request is indeed a small act which will open the door for larger actions, but I think those actions are highly desirable. So maybe we should replace the camel by something prettier in the metaphor.

DanBloomberg:
One more thing.  Following your discussion, perhaps we're actually moving toward a simple compromise where lept objects can be declared const if (1) it doesn't require casts within a function and (2) it doesn't require new 'const' functions (e.g., accessors).    And, of course, if they really are const.

DanBloomberg:
@stweil 
@jbreiden

Jeff Breidenbach asked if adding these const declarations could possibly change anything that would cause a new distribution (1.77) to require a new soname with old tesseract libraries that aren't recompiled.

In other words, if we make a new leptonica 1.77, with these const declarations, will it link properly with old tesseract executables made without them?   (I have no idea)

stweil:
Sorry for the delayed answer. The signature of C functions does not change when function arguments change, so static linking is not a problem. For shared libraries I still have to test on all relevant platforms (Linux, Windows, macOS).

PR #400, opened 2019-01-14
Add boxa2_reg and numa[12]_reg to gnuplot-dependent test list

chewi:


DanBloomberg:
Thank you James.  I'll try to remember to keep the list up to date.

DanBloomberg:
The text above says that 10 use gnuplot.  Looks like the number is now 18.  I'll change the text so that it won't need to be updated in the future.

PR #402, opened 2019-02-10
Add more const attributes to function parameters

stweil:


DanBloomberg:
I'm not ignoring this  :-)
Been out for a few days with a bad cold; should be able to look this over and merge it within a week.
  -- Dan

DanBloomberg:
Nice job.  We now have the camel's two front feet under the tent  :-)

I noticed that allheaders.h is formatted exactly as it would be if generated by xtractprotos using 'make allheaders'.  Is that how you made it?

stweil:
I changed it manually.

DanBloomberg:
It was byte-perfect.   If you make "locally" with the hand-generated makefile:
     make allheaders
 it takes about 3 seconds to generate allheaders.h from the source.

PR #419, opened 2019-04-08
implement WEBP_ANIM build for autotools and fix travis linux build.

zdenop:
- implement WEBP_ANIM build for autotools 
- fix travis linux ubuntu cmake build.


DanBloomberg:
Really fine job!!!

Just noting that in the travis logs, there are a lot of warnings about the jp2k header file, where it's defined with brackets and then redefined with quotes.

DanBloomberg:
Also, the little widget on the leptonica page has the travis build red (failing), whereas from the details it has succeeded.

zdenop:
Originally travis osx build does not used jp2k, now it is included to build. So now osx and build are comparable from point of used libraries. (Warnings comes from osx build, linux is ok...)

Difference come from point that cmake on linux does not find openjpeg header in [these check](https://github.com/DanBloomberg/leptonica/blob/9f59f921079e23e9677ebd3c017242834bcf35d8/cmake/Configure.cmake#L55), but on osx there are found, so [definition of headers follows](https://github.com/DanBloomberg/leptonica/blob/9f59f921079e23e9677ebd3c017242834bcf35d8/cmake/Configure.cmake#L148)...

I am not sure if this is needed, but I needed to have same travis configuration for osx and linux during testing this patch.

zdenop:
regarding widget: egorpugin  is [testing something](https://travis-ci.org/DanBloomberg/leptonica/builds)...

egorpugin:
It's just travis spurious fail.

DanBloomberg:
yes, I realized that there was a kind of race condition, where your travis
build was successful, but it takes a minute or two for the widget update,
and during that time Egor pushed something else, so ...

PR #421, opened 2019-04-08
Fix format strings for size_t arguments

stweil:
This fixes warnings from clang and MS VC.


DanBloomberg:
oh, I did this also

stweil:
Thank you.

PR #424, opened 2019-04-21
support tesseract autotools leptonica name

zdenop:
Create symlink (for unix like os) with autotools library name (lept)


DanBloomberg:
Thank you.

PR #395, opened 2018-12-29
add missing openjpeg libs to lept.pc

jeroen:
fixes #386

PR #429, opened 2019-05-02
add missing LIBJP2K to lept.pc; fixes #395

zdenop:
replace PR #395


DanBloomberg:
thank you both for fixing this

PR #432, opened 2019-05-09
Build improvements

zdenop:
Various improvement for cmake and autotools


DanBloomberg:
Thank you!   Does this address any of the open issues?

zdenop:
No, but now it should does not matter if you build leptonica with cmake or autotools (there are still some differences in lept.pc)

DanBloomberg:
Being able to choose between autotools and cmake on all platforms is amazing.

I really appreciate your work (with Egor, of course) on all the build issues.

PR #437, opened 2019-06-07
fixes #435 by rounding mathematically correct

### LINKED ISSUE #435: overflow in pixSauvolaGetThreshold
There is a "number overflow" in 
https://github.com/DanBloomberg/leptonica/blob/master/src/binarize.c#L766
with e.g.,
```
ms = 65024
mv = 255
ms - mv * mv = -1
```
and the `-1` is used as a negative index in 
```sd = tab[var];```

The problem is very likely here
https://github.com/DanBloomberg/leptonica/blob/master/src/convolve.c#L1245
```
ival = (l_uint32)(norm * val)
```
which should be a rounding operation instead of a cast.
In the particular case, the ```norm * val = 65024.996337283206```

vidiecan:


DanBloomberg:
I've been testing this for the past hour.  Haven't found any significant problems yet.

Issue #436, opened 2019-06-07
Failed to create symlink (cross-build, install step)


Bleach665:
Winsows 10 x64, CMake gui 3.13.4, MinGW 530_32 (from Qt lib), Android NDK r16b toolchain, rev b0519a1c6fef192cac23829daa27c621a3d4489d.

Leptonica was configured with CMake GUI and Android toolchain and built with MinGW.
Received error:
```
[ 99%] Building C object src/CMakeFiles/leptonica.dir/zlibmemstub.c.o
[100%] Linking C static library libleptonica.a
[100%] Built target leptonica
Scanning dependencies of target lept
[100%] Creating lept symlink
CMake Error: failed to create symbolic link 'liblept.so': operation not permitted
src\CMakeFiles\lept.dir\build.make:56: recipe for target 'src/CMakeFiles/lept' failed
mingw32-make.exe[2]: *** [src/CMakeFiles/lept] Error 1
CMakeFiles\Makefile2:123: recipe for target 'src/CMakeFiles/lept.dir/all' failed
mingw32-make.exe[1]: *** [src/CMakeFiles/lept.dir/all] Error 2
Makefile:128: recipe for target 'all' failed
mingw32-make.exe: *** [all] Error 2
```
The last revision which I used and which was provided successful configuring and build in this way is ad35f07eebd165580938fc1ba4c7037a43a7d2ba.

zdenop:
Thanks for report. IMO this is cmake issue: AFAIK [from this commit](https://gitlab.kitware.com/crondog/cmake/commit/afb7f6e4ff8a0c1a68c8d8cf2d6cf72401cfb8ff ) (11 moth ago) cmake should support symlinks also on windows...  Anyway I suggest to turn it off by default.

zdenop:
IMO this could closed.

DanBloomberg:
done

PR #438, opened 2019-06-09
Cmake fixes

### LINKED ISSUE #433: Wrong library name in the lept.pc
This commit https://github.com/DanBloomberg/leptonica/commit/0fefa7e55b27c1cc76e79f755a12a00c5aa3ef7d makes it impossible to use pkg-config.
So, it is ok to use a name `leptonca` here https://github.com/DanBloomberg/leptonica/commit/0fefa7e55b27c1cc76e79f755a12a00c5aa3ef7d#diff-afaa716feb616ef56fe0428c098a59abR45 because it is not a library name - it is a target name, but here https://github.com/DanBloomberg/leptonica/commit/0fefa7e55b27c1cc76e79f755a12a00c5aa3ef7d#diff-96c54e0786cbda272167018ac7bfd833R9 should be a file name.

zdenop:
- fix #433
- Workaround for #436

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@ set(MINIMUM_WEBPMUX_VERSION 0.5.0)
 
 option(BUILD_PROG "Build utility programs" OFF)
 option(CPPAN_BUILD "Build with cppan" ON)
+option(SYM_LINK "Create symlink leptonica -> lept on UNIX" OFF)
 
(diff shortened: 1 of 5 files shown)


DanBloomberg:
Thank you for fixing this, and thanks also to Aleksey-Nikolaev and Bleach665 for reporting the issues.

  -- Dan

Issue #442, opened 2019-06-30
read Jpeg and PNG produce wrong alpha in 32bpp pix


alexcohn:
In [tesseract 4 Android](https://github.com/rmtheis/tess-two) (and its multiple forks, including a recent [rewrite](https://github.com/adaptech-cz/Tesseract4Android)), we use **pix** with 32bpp.

It happens so that some Java tests written for image file i/o fail, because the underlying **pngio.c** and **jpegio.c** don't set the **L_ALPHA_CHANNEL** byte to **255**. For **JPG** this is unconditional; for **PNG**, when **spp == 3**.

DanBloomberg:
There are two issues (png and jpeg).

png: would it solve the problems if when writing 3 spp png, do the same, set the alpha sample to 255?

jpeg: I don't understand the problem, because jpeg compressed files do not have an alpha channel; they have spp = 1 or 3, only.

DanBloomberg:
I see, the failure is on the pix after reading -- you want the alpha byte to be set to 255 for all spp =3 pix.

alexcohn:
Yup.

alexcohn:
I don't think you *can* set alpha to 255 when the PNG is 3spp, it's ***assumed*** that in this case, opacity is full.

alexcohn:
The only question is, how this failing test has triggered nobody before. Well, some WriteFile tests in tesseract for Android fail, too; but I chose to begin with read.

DanBloomberg:
We agree, the fix would be in reading a pix.  But really, the issue is what the test does with that pix.

I would guess that the test is flaky.  We have an spp == 3 pix, and the assumption should be that the 4th byte in the pix can be anything.  It should never be tested.

alexcohn:
OK, this definitely explains how such thing passed under the radar. The Java wrapper at https://github.com/rmtheis/tess-two/blob/master/tess-two/src/com/googlecode/leptonica/android/Pix.java does not expose **Pix.spp**, thus the tests fail.

DanBloomberg:
fix the java tests.  See my last comment in #443

alexcohn:
I have fixed the Java tests in [my fork](https://github.com/alexcohn/Tesseract4Android). of Tesseract 4 Android. For that purpose, I have exposed **spp** and **imageFormat** to Java.

DanBloomberg:
resolved in leptonica

PR #443, opened 2019-06-30
Fix alpha byte for 32bpp pix

alexcohn:
fixing https://github.com/DanBloomberg/leptonica/issues/442

### DIFF
--- a/src/pngio.c
+++ b/src/pngio.c
@@ -332,7 +332,7 @@ PIXCMAP     *cmap;
                 SET_DATA_BYTE(ppixel, COLOR_BLUE, rowptr[k++]);
                 if (spp == 4)
                     SET_DATA_BYTE(ppixel, L_ALPHA_CHANNEL, rowptr[k++]);
-                else /* spp == 4 */
+                else /* spp == 3 */
                     SET_DATA_BYTE(ppixel, L_ALPHA_CHANNEL, 255);
                 ppixel++;
             }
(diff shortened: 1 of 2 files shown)


src/pngio.c
```diff
@@ -330,7 +330,10 @@ PIXCMAP     *cmap;
                 SET_DATA_BYTE(ppixel, COLOR_RED, rowptr[k++]);
                 SET_DATA_BYTE(ppixel, COLOR_GREEN, rowptr[k++]);
                 SET_DATA_BYTE(ppixel, COLOR_BLUE, rowptr[k++]);
-                SET_DATA_BYTE(ppixel, L_ALPHA_CHANNEL, (spp == 4) ? rowptr[k++] : 255);
+                if (spp == 4)
```
rmtheis: Consider `spp == 3`. Either the condition is wrong or the comment on line 335 is wrong

DanBloomberg:
We understand that the problem is with the java test, not the leptonica readers.

So fix the tesseract tests, and in a week when I get back from a short vacation I'll modify all the leptonica RGB readers to store 255 in the 4th byte, and also comment that no program should use the 4th byte when spp == 3

alexcohn:
Note that I have fixed the Java tests, but this fix may be helpful also in other situations.

Robyer:
To make this better you may update also code for BMP and possibly other formats to make this behavior consistent over whole library.

For BMP format add this on line 313 in file `bmpio.c`:
`*((l_uint8 *)pword + L_ALPHA_CHANNEL) = 255;`

DanBloomberg:
yes, will look at all read functions on RGB images

PR #444, opened 2019-07-12
Avoid error messages printed when using pixGenTextblockMask in pixGetRegionsBinary

vidiecan:
Problem: 
Error messages printed out when `pixGenTextblockMask` is called on a non empty page.

Description:
In the function `pixGenTextblockMask` called from `pixGetRegionsBinary`, a mask is created by calling
```
pix1 = pixMorphSequence(pixs, "c1.10 + o4.1", 0);
```
However, the mask can be empty which would result in several error messages printed out because e.g., `pixMorphSequenceByComponent`->`pixaMorphSequenceByComponent` expects non empty arrays, hence returns error and it has a cascading effect. 

Possible solution:
This PR checks if the mask in `pixGenTextblockMask` is not empty and if positive it outputs an info message instead. Minor changes in `pixGetRegionsBinary` handle the empty result.


DanBloomberg:
Thank you for doing this.  You indeed fixed the issue.  I just put the changes in, a bit differently.

#435 was fixed a month ago, so I believe all the issues here are resolved.

DanBloomberg:
I believe this is resolved.

PR #445, opened 2019-07-16
fixing https://github.com/rmtheis/tess-two/issues/159

alexcohn:
The issue with the Java test is that **pixaReplacePix()** does not increase `pix->refcount`. Therefore, destroying both the **pix** and the **pixa** reaches `refcount == -1`, and results in a crash.

### DIFF
--- a/src/pixabasic.c
+++ b/src/pixabasic.c
@@ -1315,6 +1315,7 @@ BOXA  *boxa;
 
     pixDestroy(&(pixa->pix[index]));
     pixa->pix[index] = pix;
+    pixChangeRefCount(pix, 1);
 
     if (box) {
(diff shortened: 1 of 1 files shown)


DanBloomberg:
Not a bug.  The pixa owns the new pix.  Don't destroy both.  See my other comment on this.

alexcohn:
Maybe it would be prudent to add (debug only) check that the refcount never goes below 0 in destroy functions?

Such check is irrelevant for release build, because this would access memory that has already been freed, so even if you an still read refcount, it will be an arbitrary number. OTOH, in debug, most systems retain freed memory, and this value can be quite reliably verified.

PR #456, opened 2019-11-01
Use JPEG_LIBRARIES instead of JPEG_LIBRARY

theirix:
Plural variables are always defined while singular variables are obsolete and can be removed
Detecting using _LIBRARY instead of _LIBRARIES can broke autogenerated cmake files that do not include obsolete variables.

The second commit enables the use of ZLIB_INCLUDE_DIRS that is only documented variable defined by FindZLIB module.

However, it is not possible to use *_INCLUDE_DIRS uniformly for other cases (such as TIFF, PNG) because they were introduced rather recently in CMake (3.12.0 in 2018).

Checked all modules for an ancient cmake 2.8.12 and modern 3.15.


DanBloomberg:
@egorpugin 

Should this be pushed?

egorpugin:
Sure! Thank you!

Issue #454, opened 2019-10-28
Test suite fails without external image libraries


iskunk:
I am building Leptonica for use in a software system where image I/O and compression are handled by other components. Therefore, I configure with
```
--without-zlib
--without-libpng
--without-jpeg
--without-giflib
--without-libtiff
--without-libwebp
--without-libopenjpeg
```

When I build in this manner, however, `make check` gives rather discouraging results:
```
FAIL: adaptmap_reg
FAIL: adaptnorm_reg
FAIL: affine_reg
FAIL: alphaops_reg
FAIL: alphaxform_reg
SKIP: baseline_reg
FAIL: bilateral2_reg
FAIL: bilinear_reg
FAIL: binarize_reg
[...]
============================================================================
Testsuite summary for leptonica 1.78.0
============================================================================
# TOTAL: 125
# PASS:  3
# SKIP:  18
# XFAIL: 0
# FAIL:  104
# XPASS: 0
# ERROR: 0
```

The `adaptmap_reg.log` file begins with
```

////////////////////////////////////////////////
////////////////   adaptmap_reg   ///////////////
////////////////////////////////////////////////
leptonica-1.78.0 : (null)
Error in pixReadStreamJpeg: function not present
Error in pixReadStream: jpeg: no pix returned
Error in pixRead: pix not read
Error in pixConvertRGBToGray: pixs not defined
Error in pixaAddPix: pix not defined
[...]
```

I think it would make more sense for the image files that are used in the library's test suite to be distributed in a format that the library can read without external dependencies, such as bmp or pnm. (Of course, this would not apply in instances where I/O involving formats that require external dependencies is the functionality being tested.)

DanBloomberg:
If I were to have all the image files that are used for testing in uncompressed format, they would take up about 15 times the current disk space, which I've been careful to keep down to about 10 MB.  The approx. 200 jpg, png and tiff compressed image files would balloon to an estimated 150 MB.

iskunk:
As I understand, bmp format still does RLE, and the image files would be compressed within the distribution tarball (or .zip file) in any event. Would that improve the situation much?

Another possibility: add a set of Make rules that use some external tool (e.g. ImageMagick's `convert` utility) to convert the image files into bmp/pnm, combined with a fallback in the code to load in such a format if the normal jpg/png/tiff support is unavailable.

DanBloomberg:
It doesn't make much sense to me.

[...]

PR #457, opened 2019-11-08
Fix several build problems and potential bugs

stweil:
Fix clang and gcc warnings, one of them indicating a buggy `for` statement.

### DIFF
--- a/prog/baseline_reg.c
+++ b/prog/baseline_reg.c
@@ -35,6 +35,10 @@
  *       mottled text image.
  */
 
+#ifdef HAVE_CONFIG_H
+#include <config_auto.h>
+#endif /* HAVE_CONFIG_H */
+
 #include "allheaders.h"
 
 int main(int    argc,
@@ -46,7 +50,7 @@ PIXA         *pixadb;
 PTA          *pta;
 L_REGPARAMS  *rp;
 
-#if !defined(HAVE_LIBPNG)
+#if !HAVE_LIBPNG
     L_ERROR("This test requires libpng to run.\n", "baseline_reg");
     exit(77);
 #endif
--- a/prog/colormask_reg.c
+++ b/prog/colormask_reg.c
@@ -32,6 +32,10 @@
  *   peaks in HS.
  */
 
+#ifdef HAVE_CONFIG_H
+#include <config_auto.h>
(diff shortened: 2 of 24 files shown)


stweil:
@DanBloomberg, I'll add more commits, but you can merge them at any time.

DanBloomberg:
pixarith.c:  Wow.  I was just trying to write a test for this function, which had failed.  You solved that for me!

sel1.c:   fine

writefile.c:   do we need lines 91-93, once you've removed 'defined' ?

fix undefined macros in conditional compilations:

  for those with HAVE_FMEMOPEN, can't we just add to l. 171 in environ.h:
#else 
#define HAVE_FMEMOPEN   0

  for fpix2.c, just comment those two lines out
   /*    for debugging
    two lines
   */

  for readbarcode.c, can add
   #define DEBUG_FREQUENCY at the top

unreachable break statements:  fine

DanBloomberg:
For fix prototypes for functions without header: fix by adding 'void' to the function definitions.

boxa3_reg:  fine

You have another one but I can't find it now!

DanBloomberg:
fix preprocessor conditionals:  fine

DanBloomberg:
Maybe get these ready for a commit before adding more?

DanBloomberg:
Let me do the 'fix functions without a parameter"  -- adding void to the definition

DanBloomberg:
I'm confused about :
#ifdef HAVE_CONFIG_H
￼#include <config_auto.h>
￼#endif /* HAVE_CONFIG_H */

Do we need this at all?  How does config_auto.h get included normally?
Of if we need it in some/all of the source files, shouldn't it be in environ.h?

DanBloomberg:
I now believe that it's important to add:

#ifdef HAVE_CONFIG_H
￼#include <config_auto.h>
￼#endif /* HAVE_CONFIG_H */

to all the source files, and I will try to do this over the weekend.
Also, I pushed the change in pixarith.c that you found.

stweil:
I created now separate pull requests for some commits, so it is easier for you to merge them.

stweil:
Now the remaining commit are obsolete because they will be replaced by Autoconf conforming code (see discussion for issue #454). I therefore close this pull request.

PR #462, opened 2019-11-09
Fix checks which require gnuplot

stweil:
Those checks were skipped since commit 1b3c9b64ee204b3f9371cea624c959b3177c9c1d
because the `sh` shell does not support `type -P`.


DanBloomberg:
Merged.  Did you check with automake to validate?
(the travis builds use cmake)

stweil:
Yes, I used automake and `make check` (it's part of my script for Coverity Scan).

DanBloomberg:
ah, just read this.  Great!

DanBloomberg:
I merged and asked before seeing your comment on the PR     :-)

PR #463, opened 2019-11-09
Fix writeCustomTiffTags for type `const char*`

stweil:
This fixes mtiff_reg. Add also a message for an unknown type.


DanBloomberg:
Thanks!

PR #467, opened 2019-12-03
CMake: Use BUILD_SHARED_LIBRARIES and get rid of LIBRARY_TYPE

1480c1:
Still retain `STATIC` for backwards compatibility, but just map `!STATIC` to setting [`BUILD_SHARED_LIBRARIES`](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html)

If wanted, I can reverse the mapping, but I don't see a reason to do that since [`BUILD_SHARED_LIBRARIES`](https://cmake.org/cmake/help/v2.8.11/cmake.html#variable:BUILD_SHARED_LIBS) exists even on 2.8.11


DanBloomberg:
@egorpugin 

please eval

egorpugin:
@1480c1 thank you.
Could you please remove STATIC completely from cmake files? And option()s you've introduced too.
And use BUILD_SHARED_LIBRARIES instead.

1480c1:
> @1480c1 thank you.
> Could you please remove STATIC completely from cmake files? And option()s you've introduced too.
> And use BUILD_SHARED_LIBRARIES instead.

Sure, I mainly kept `STATIC` since that was already there

egorpugin:
There are also two STATIC checks in src/cmakelists and prog/cmakelists.

egorpugin:
Thanks!

1480c1:
Thank you

PR #469, opened 2019-12-04
CMake: Only expose SYM_LINK if UNIX

1480c1:
`SYM_LINK` only has an effect on UNIX since the code that depends on it is guarded by `if(UNIX)` so it would not make sense to expose it to windows users

It may be better to instead add a check to see if `cmake -E create_symlink` works and then if `SYM_LINK` is on, add the `lept` target to `ln` the library since, as mentioned in https://github.com/DanBloomberg/leptonica/issues/436, cmake for windows also gained the `create_symlink` command, however, it is still not generally available, requires running the shell as admin, or developer mode on windows


egorpugin:
Please, mention me in any other cmake PRs, so I react faster. :)

1480c1:
Okay, will do

Issue #472, opened 2020-01-04
Several tests failing in 1.79.0


chewi:
`pixa1_reg` fails because it's not gated on gnuplot. Please check that before releasing, I've lost count of the number of times I've had to fix it now. `rank_reg` also logs the same error but seems to pass anyway.

I don't know why the other 3 are failing.

```
////////////////////////////////////////////////
////////////////   webpio_reg   ///////////////
////////////////////////////////////////////////
leptonica-1.79.0 : libgif 5.2.1 : libjpeg 6b (libjpeg-turbo 2.0.3) : libpng 1.6.37+apng : libtiff 4.1.0 : zlib 1.2.11 : libwebp 1.0.3
Time to read jpg:   0.000
Time to write webp:   0.002
Time to read webp:   0.000
Time to read jpg:   0.000
Time to write webp:   0.002
Time to read webp:   0.000
Time to read jpg:   0.001
Time to write webp:   0.027
Time to read webp:   0.003
Time to read jpg:   0.011
Time to write webp:   0.219
Error in fopenReadStream: file not found
Error in l_binaryRead: file stream not opened
Error in fileCopy: data not returned
Error in fopenReadStream: file not found
Error in pixRead: image file not found: /tmp/lept/webp/webpio.6.webp
Time to read webp:   0.000
Error in regTestCompareSimilarPix: pix1 and pix2 not both defined
Error in fopenReadStream: file not found
Error in pixRead: image file not found: /tmp/lept/webp/junk.webp
Error in pixGetPSNR: empty input pix
qual = 50, psnr =   0.000
Failure in webpio_reg: value comparison for index 7
difference = 43.500000 but allowed delta = 1.000000
qual = 75, psnr =  46.106
qual = 90, psnr =  52.401
qual = 100, psnr =  57.460
Lossless write:   0.763 sec
lossless; psnr should be 1000: psnr = 1000.000
Time:   2.173 sec

////////////////////////////////////////////////
////////////////   pdfio1_reg   ///////////////
////////////////////////////////////////////////
leptonica-1.79.0 : libgif 5.2.1 : libjpeg 6b (libjpeg-turbo 2.0.3) : libpng 1.6.37+apng : libtiff 4.1.0 : zlib 1.2.11 : libwebp 1.0.3

*** Writing single images as pdf files

*** Writing multiple images as single page pdf files

*** Writing segmented images without image regions
Warning in pixGenerateCIData: pixs has < 8 bpp; using flate encoding
[...]

PR #478, opened 2020-03-24
Fix parallel running of tests by using unique directories

chewi:
Following #472, I did find a way to make some tests depend on others but I then found that the only reason they break in parallel is because some tests write to the same directories for seemingly no reason. `pdfio2_reg` also depends on a couple of files written by `pdfio1_reg` but it's only a few small extra lines to make it also generate the same files. All the other tests are independent so it seems silly to have this one exception.


DanBloomberg:
Thank you for fixing these problems.

PR #480, opened 2020-04-26
Move array lengths to #defines.

robinwatts:
Using "static const int" for array lengths is a C feature too
new for Visual Studio's C compiler. Move to using enums instead
as this compiles everywhere and has no functional difference.

### DIFF
--- a/src/classapp.c
+++ b/src/classapp.c
@@ -56,7 +56,7 @@
 #include <string.h>
 #include "allheaders.h"
 
-static const l_int32 L_BUF_SIZE = 512;        /*!< size of filename buffer    */
+enum { L_BUF_SIZE = 512 };                    /*!< size of filename buffer    */
 static const l_int32 JB_WORDS_MIN_WIDTH = 5;  /*!< min. word width in pixels  */
 static const l_int32 JB_WORDS_MIN_HEIGHT = 3; /*!< min. word height in pixels */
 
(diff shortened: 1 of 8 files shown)


src/classapp.c
```diff
@@ -56,7 +56,7 @@
 #include <string.h>
 #include "allheaders.h"
 
-static const l_int32 L_BUF_SIZE = 512;        /*!< size of filename buffer    */
+enum { L_BUF_SIZE = 512 };                    /*!< size of filename buffer    */
```
stweil, line 59: I'd use `#define BUF_SIZE 512` then like it is already done in several other files and rename the macro name everywhere to match that definition. Using an enum looks very artificial.

stweil:
Would `static const L_BUF_SIZE = 512;` work for the MS compiler?

src/classapp.c
robinwatts: I would absolutely be happy to use #define if that's preferred.

#define's are source transformations at heart, so are a step further away from static const's than an enum is, but if that's the preference of the maintainer, so be it.

robinwatts:
> Would `static const L_BUF_SIZE = 512;` work for the MS compiler?

No. Cos that's that same thing. Omitting the 'int' merely makes it an int by default.

robinwatts:
I'll reopen with a different commit so the tests rerun.

PR #481, opened 2020-04-26
Move array lengths to #defines.

robinwatts:
Using "static const int" for array lengths is a C feature too
new for Visual Studio's C compiler. Move to using #defines instead
as this compiles everywhere and has no functional difference.

### DIFF
--- a/src/classapp.c
+++ b/src/classapp.c
@@ -56,7 +56,7 @@
 #include <string.h>
 #include "allheaders.h"
 
-static const l_int32 L_BUF_SIZE = 512;        /*!< size of filename buffer    */
+#define L_BUF_SIZE 512                        /*!< size of filename buffer    */
 static const l_int32 JB_WORDS_MIN_WIDTH = 5;  /*!< min. word width in pixels  */
 static const l_int32 JB_WORDS_MIN_HEIGHT = 3; /*!< min. word height in pixels */
 
--- a/src/fhmtauto.c
+++ b/src/fhmtauto.c
@@ -103,7 +103,7 @@
 
 #define   PROTOARGS   "(l_uint32 *, l_int32, l_int32, l_int32, l_uint32 *, l_int32);"
 
-static const l_int32  L_BUF_SIZE = 512;
+#define L_BUF_SIZE 512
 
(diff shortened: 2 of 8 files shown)


stweil:
The current code is more than 4 years old, so it looks like you are the first one to use the code with MS C. Did you get it running, or are more changes needed?

zdenop:
Please provide information how do you try to build leptonica, VS version, exact error message etc.

egorpugin:
Looks like this is leptonica build as C files.
Proper way currently is to build .c files as C++.

DanBloomberg:
It seems that VS C doesn't support C99.  (21 years later)
Has this not come up in the past few years because everyone with VS uses the C++ compiler?

egorpugin:
VS2019 16.6 Preview has C99 conformance. Will be released in a month or two.
16.5 is the latest release now.

robinwatts:
> The current code is more than 4 years old, so it looks like you are the first one to use the code
> with MS C. Did you get it running, or are more changes needed?

Yes, I have Leptonica working under Visual C, both using a home-made VS solution and home made nmake files.

robinwatts:
Regardless, using variable-length arrays incurs a runtime cost. We don't need the variable-length nature, so why take that hit?

I am investigating using leptonica/tesseract within Ghostscript (and family). GS compiles on *everything*, so it'd be really nice not to introduce any new wrinkles where we don't have to. Having to support variable-length arrays (or get .c files built as c++ ones) are both wrinkles we don't need.

I mean, if you won't take the tiny patch, we'll have to, but it seems a shame.

DanBloomberg:
It's not a big deal.  I changed these variables from #define to static const ... 4 or 5 years ago, and we should go back for maximum compatibility.

egorpugin:
I've added more fixes in https://github.com/DanBloomberg/leptonica/commit/533fa112305a71c90f1ab361b986101d793db6e2

robinwatts:
Thanks!

amitdo:
@robinwatts commented on Apr 27

>I am investigating using leptonica/tesseract within Ghostscript (and family).

https://www.ghostscript.com/ocr.html

DanBloomberg:
Noted.  Embedding tesseract OCR in ghostscript is an excellent next step.

PR #483, opened 2020-04-28
Moved over fuzzers and corpus from oss-fuzz

AdamKorcz:
I am hereby moving over the current state of fuzzers and corpus from oss-fuzz to Leptonicas own repository as suggested by Abhishek (https://github.com/google/oss-fuzz/pull/3712#issuecomment-619474561)

This PR does not add any new functionality to the fuzzers or the corpus. There have been made 2 insignificant changes compared to the files over on oss-fuzz's repository:

1. The license header has been removed from the fuzzers as it is not required when the fuzzers are hosted on the upstream repository.
2. The name of "corpus.zip" has been changed to "general_corpus.zip".

The files are submitted to `/prog/fuzzing` since the rest of the tests are in `/prog`, but any alternative should be fine in terms of running the fuzzers.

Once the files have been accepted here, I will apply the necessary changes to the build-process on oss-fuzz, as they will need to be accepted first to pass the Travis-tests.

@stweil had a great suggestion to reducing the effort in the build-script (https://github.com/google/oss-fuzz/pull/3712#issuecomment-619520917), which I will be happy to implement once we get the fuzzers up and running. Likewise, if you should have any comments or suggestions to the current fuzzers, I will be happy to hear them and submit them in a seperate PR to not mix PRs together.

The fuzzers are currently failing on oss-fuzz, and we will get them up and running again with this setup.


DanBloomberg:
Why put these fuzzers in leptonica, and not in a directory of oss-fuzz?
Aren't most of the oss fuzzers located in project oss-fuzz?

Also, I do know that the leptonica fuzzers need to be built with the flag
  FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION

@stweil    Why are you suggesting this?
@inferno-chromium  I don't  know who to ask in the oss-fuzz project.

inferno-chromium:
See why
https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/#fuzz-target
"This makes it easy to maintain the fuzzers and minimizes breakages that can arise as source code changes over time."

DanBloomberg:
OK, let's give it a try.

stweil:
> Aren't most of the oss fuzzers located in project oss-fuzz?

Like the other examples shown for ideal integration, Tesseract also has the fuzzer code in its own codebase: https://github.com/tesseract-ocr/tesseract/tree/master/unittest/fuzzers. I think it is helpful for Leptonica, too.

PR #484, opened 2020-04-29
Added script to build corpora of fuzzers

AdamKorcz:
I am adding this script so that managing the corpora is done on Leptonicas own repository.

The file will be executed on line 150 in the build script (https://github.com/google/oss-fuzz/blob/master/projects/leptonica/build.sh), and the commands currently at line 150 will be removed from oss-fuzz's repository. 

It gives the maintainers of Leptonica more freedom to update the corpora by not having to wait for the oss-fuzz team to merge the updates in.

I will update the build script on oss-fuzz and rewrite the compilation of the fuzzers so that part doesn't have to be updated whenever a new fuzzer is added. This should remove all need to make PR's to oss-fuzz when a new fuzzer is added, modified or removed.

### DIFF
--- a/prog/fuzzing/build_corpora.sh
+++ b/prog/fuzzing/build_corpora.sh
@@ -1,0 +1,4 @@
+cp $SRC/leptonica/prog/fuzzing/general_corpus.zip $OUT/pix_rotate_shear_fuzzer_seed_corpus.zip
+cp $SRC/leptonica/prog/fuzzing/general_corpus.zip $OUT/enhance_fuzzer_seed_corpus.zip
+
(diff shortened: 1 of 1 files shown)


stweil:
Wouldn't it be reasonable to move most of the build script to Leptonica (as it is done for [LibreOffice](https://github.com/google/oss-fuzz/blob/master/projects/libreoffice/build.sh))?

AdamKorcz:
I am open to that, and yes, it would make things much simpler on oss-fuzz's end. 

Shall we make a move for that instead?

stweil:
I think that would be a good solution.

PR #485, opened 2020-04-29
Added build script from oss-fuzz

AdamKorcz:
I am adding the build-script from oss-fuzz.

The script will be executed like with [LibreOffice ](https://github.com/google/oss-fuzz/blob/master/projects/libreoffice/build.sh)in build.sh in oss-fuzz.

Having the build script on Leptonicas own repository gives the maintainers of Leptonica more freedom to update the fuzzers and corpora by not having to wait for the oss-fuzz team to merge the updates in.

This PR adds 2 modifications to the build script compared to the current version on oss-fuzz:

1. The license header has been removed as it is not needed here.
2. The compilation of the fuzzers has been rewritten to prevent having to update the build script every time a fuzzer is added or removed. As long as each fuzzer is named *_fuzzer.cc and is placed in the `/prog/fuzzing` directory, it will be included.

### DIFF
--- a/prog/fuzzing/oss-fuzz-build.sh
+++ b/prog/fuzzing/oss-fuzz-build.sh
@@ -10,14 +10,6 @@ pushd $SRC/zstd
 make -j$(nproc) install PREFIX="$WORK"
 popd
 
-# libjbig
-pushd "$SRC/jbigkit"
-make clean
-make -j$(nproc) lib
-cp "$SRC"/jbigkit/libjbig/*.a "$WORK/lib/"
-cp "$SRC"/jbigkit/libjbig/*.h "$WORK/include/"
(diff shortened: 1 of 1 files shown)


DanBloomberg:
Is libjbig necessary?  It is not used in leptonica.

Is the build command for the executable defined elsewhere?  I'm thinking about the fuzzing variable FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION that needs to be defined at compile time.

timothy-king:
libjbig: I don't remember if it is necessary. If the integration with oss-fuzz still [works](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#testing-locally), I would not worry about it.

PR #494, opened 2020-05-01
Simplified pixa_recog_fuzzer

AdamKorcz:
This PR removes the loop in the fuzzer that causes a null-dereferencer.

### DIFF
--- a/prog/fuzzing/pixa_recog_fuzzer.cc
+++ b/prog/fuzzing/pixa_recog_fuzzer.cc
@@ -9,52 +9,39 @@
 extern "C" int
 LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
-        if(size<20) return 0;
+        if(size<10) return 0;
 
         char filename[256];
         sprintf(filename, "/tmp/libfuzzer.pa");
-
         FILE *fp = fopen(filename, "wb");
-        if (!fp)
-                return 0;
+        if (!fp) return 0;
         fwrite(data, size, 1, fp);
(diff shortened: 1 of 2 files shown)


prog/fuzzing/pixa_recog_fuzzer.cc
```diff
@@ -9,52 +9,39 @@
...
-        l_int32    histo[10];
         PIXA      *pixa1, *pixa2, *pixa3, *pixa4;
-        L_RECOG   *recog1;
-        l_int32    i, n, ival;
-        PIX       *pix1;
+        L_RECOG   *recog1, *recog2;
+        PIX       *pix1, *pix2;
 
         pixa1 = pixaRead(filename);
         pixa2 = pixaCreate(0);
```
DanBloomberg, line 26: you made two pixa2.  The first one (l. 26) isn't used.  
pixa4 and pix1 and pix2 are not used.  Remove.  Deleting would crash because they are non-zero garbage.

DanBloomberg: I meant that use of pixDestroy() etc would cause a crash.

AdamKorcz:
An unrelated commit sneaked into this one. I apologize for the inconvenience. I am closing here and opening a new one with two seperate ones.

stweil:
You could also remove the unwanted commit locally and make a forced push to fix such situations.

AdamKorcz:
Ah sure of course. Thanks for the heads up.

PR #497, opened 2020-05-02
Improve fuzzers and add basic documentation and rules for local fuzzing

stweil:


DanBloomberg:
This looks very good.  I have one question out of ignorance.  If we are building the fuzzers in leptonica instead of in oss-fuzz (or google/oss-fuzz), don't we need the compiler flag FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION?

You can see how it is used in pix1.c When this is defined, the maximum size of the images that can be read is reduced to avoid timeouts and out-of-memory errors.

stweil:
It can be set if desired by the usual mechanisms when running `configure`, but I don't think it is required for fuzzers which are running locally, not in the oss-fuzz infrastructure.

stweil:
While thinking twice about it, I now think that I might be wrong. But we can add that compiler flag any time if we see that it is needed.

DanBloomberg:
OK.  I was plagued with these timeout and OOM fuzzer errors for a long time before we realized we could use that variable to avoid them.

PR #500, opened 2020-05-05
Make reading of PNM files more robust

stweil:
Programs that read this format should be as lenient as possible,
accepting anything that looks remotely like a pixmap.
See http://www.fileformat.info/format/pbm/egff.htm.

As other software (like for example convert) seems to ignore
non-digits in the input stream, do it here, too.

### DIFF
--- a/src/pnmio.c
+++ b/src/pnmio.c
@@ -1346,10 +1346,10 @@ l_int32   c, ignore;
     *pval = 0;
     if (!fp)
         return ERROR_INT("stream not open", procName, 1);
-    do {  /* skip whitespace */
+    do {  /* skip whitespace and non-numeric characters */
         if ((c = fgetc(fp)) == EOF)
             return 1;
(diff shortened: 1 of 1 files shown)


stweil:
This commit also fixes the timeout in oss-fuzz:

The test data included an '&' character, so fscanf did not get a number, the file read position was not advanced, and the same (wrong) data was read again and again, never reaching EOF.

With the new code, images with large width and height will be read very fast because the fuzzing data is not very large, so EOF is reached fast.

stweil:
The old code also missed to check for all white-space characters. Extract from `man isspace`:

    In the "C" and "POSIX" locales, these are: space, form-feed ('\f'), newline ('\n'), 
    carriage  return  ('\r'),  horizontal  tab ('\t'), and vertical tab ('\v').

Maybe other locations in the Leptonica code should use `isspace`.

DanBloomberg:
Thank you.   This indeed solves the problems with fuzzing pnm.

PR #501, opened 2020-05-05
Added seed corpus for barcode fuzzer

AdamKorcz:
This PR adds a seed corpus for the barcode fuzzer and adds a line to move it to `/out/` in the oss-fuzz build file where it will be read by oss-fuzz.

The zip contains the .png and .jpg barcode-testfiles from `leptonica/prog/`


stweil:
Would symbolic links instead of copies be sufficient, too?

AdamKorcz:
In the case of oss-fuzz I would not recommend it due to the use of Docker images. Everything that the fuzzers need has to be placed in `/out/`. 
The current setup is recommended in [the oss-fuzz documentation](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#seed-corpus).

PR #503, opened 2020-05-06
Added more targets to colorquant_fuzzer

AdamKorcz:
This PR adds more coverage to the colorquant fuzzer. The fuzzer is meant to keep the targets in the the colorquant files inside a single fuzzer. If it grows any bigger, we may need to split it up.


DanBloomberg:
looks good

PR #506, opened 2020-05-07
Added new fuzzer

AdamKorcz:
This PR adds a new fuzzer that targets `ccbord.c`.

A small note on this fuzzer:
It runs out of memory fairly quickly, and it seems a leak is coming from `ccbaStepChainsToPixCoords()`.

It seems to me that everything is cleaned up the way it should be, and I appreciate any comments if that isn't the case.

### DIFF
--- a/prog/fuzzing/ccbord_fuzzer.cc
+++ b/prog/fuzzing/ccbord_fuzzer.cc
@@ -1,0 +1,27 @@
+#include "leptfuzz.h"
+
+extern "C" int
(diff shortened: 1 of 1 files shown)


DanBloomberg:
I'll look for leaks in that function.

DanBloomberg:
There are several places for memory leaks, that happen with an early return.
I'll fix them.

DanBloomberg:
I believe the leaks are now plugged.

PR #509, opened 2020-05-08
Added new fuzzer

AdamKorcz:
This PR adds a fuzzer that targets mask functionality of Leptonica.


DanBloomberg:
in line 35, pix2 hasn't been defined (it's garbage on the stack).  It needs to be a 8 or 32 bpp pix.

DanBloomberg:
this will always fail.  pix1 must be 1 bpp; pix2 must be 8 or 32 bpp.

perhaps do
pix2 = pixConvertTo8(pix1, FALSE);

PR #512, opened 2020-05-11
Used pixCopy on pixs2 instead of pixReadMemSpix

AdamKorcz:
Currently the fuzzer does not go past pixs2. This PR suggests a fix for that.


DanBloomberg:
I am surprised that read function would fail the second time.  I need to investigate.
For now, this should solve the problem.

DanBloomberg:
Checked & also tested.  pixReadMemSpix() is idempotent for all types of images.
Perhaps the fuzzer is somehow grabbing different data every time it reads from that buffer ???

PR #513, opened 2020-05-11
Added new fuzzer

AdamKorcz:
This PR adds a fuzzer that targets `boxaReconcileAllByMedian()`


amitdo:
There is an oss-fuzz badge. If you are interested in adding one, see:    

https://github.com/tesseract-ocr/tesseract

DanBloomberg:
@amitdo 
Good to know about the badge.  We should probably add it in leptonica as well.

PR #514, opened 2020-05-12
Avoid brokenness with custom pix memory handlers.

robinwatts:
setPixMemoryHandler allows custom malloc/free handlers to be
set. Then the code goes and uses calloc to get a block that
it later feeds to the custom free handler. Obviously, this
leads to crashes.

### DIFF
--- a/src/pix1.c
+++ b/src/pix1.c
@@ -244,6 +244,27 @@ pix_malloc(size_t  size)
 #endif  /* _MSC_VER */
 }
 
+static void *
+pix_calloc(size_t count, size_t size)
+{
+    char *block;
+    size_t total = count * size;
+
+    /* Check for overflow */
+    if (count == 0 || size == 0 || total / count != size)
+        return NULL;
+#ifndef _MSC_VER
+    block = (*pix_mem_manager.allocator)(total);
+#else  /* _MSC_VER */
+    /* Under MSVC++, pix_mem_manager is initialized after a call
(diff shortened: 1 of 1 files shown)


robinwatts:
I need to find some way to solve the general case (i.e. ALL allocations, on all platforms, not just the pix ones), but at least with this commit the defined interface won't cause crashes.

DanBloomberg:
Thank you for submitting this.

It was never intended to do memory management on anything besides images.
The reason is that they can be very large, and an application may be improved by not freeing and reallocating many if they're the same size, for example.

Allocators for small objects are very efficient, and memory fragmentation isn't usually an issue.

DanBloomberg:
I don't understand what you did to get a crash.
The custom allocator is only used for the data array in the pix.
All other data arrays, and all structs like the Pix are allocated with the system malloc and alloc.

Your PR shows in lines 555 and 667 the use of the custom allocator to calloc and free the Pix struct itself, not the raster data.    Why was it necessary to do that?

robinwatts:
On Monday, I had absolutely convinced myself that this was required. I had crashes in my test code to indicate it. First I had a crash in the free which lead to my "fix", which lead (obviously) to problems with the calloc.

Then yesterday, I found a bug in my code with interaction with garbage collection.

Looking at the changes again now, especially in light of your comments, I can't see why these changes could be necessary. I can only assume that I misdiagnosed the initial problem I found given a crash on free caused by the gc issue.

So, sorry, apologies for having wasted your time!

PR #515, opened 2020-05-12
[pix4.c] Added fuzzer

AdamKorcz:
This PR adds a fuzzer that targets the functions in pix4.c that need to be explicitly called in order to be covered.

### DIFF
--- a/prog/fuzzing/pix4_fuzzer.cc
+++ b/prog/fuzzing/pix4_fuzzer.cc
@@ -28,7 +28,8 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 
     pix1 = pixConvertTo8(pixs, FALSE);
     box = boxCreate(120, 30, 200, 200);
-    pixGetCmapHistogramInRect(pix2, box, 1);
+    numa1 = pixGetCmapHistogramInRect(pix2, box, 1);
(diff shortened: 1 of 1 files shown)


AdamKorcz:
Thanks a lot @DanBloomberg

AdamKorcz:
Thank you for the comments. 
Regarding line 94, it did return a pix before, but the name of the variable itself was confusing, so I have changed it.

DanBloomberg:
@AdamKorcz 

3 questions:
(1) Is github testing these fuzzer programs to see if they compile?
(2) How can I compile them?
(3) fuzzing/oss-fuzz-build.sh has a section at the end that copies zip files from $SRC to $OUT.  What's happening here?  I only see 10 of the fuzzers there, but we have 17 fuzzers.

AdamKorcz:
@DanBloomberg 
1: No, Github is not testing these fuzzers. If they fail, we will currently only get notified through oss-fuzz. I did run this fuzzer for 30 minutes with a seed corpus, and it did not break. It would be a great idea to add the fuzzers to the automated checks here on Github, though.
2: I think @stweil setup documentation for running the fuzzers locally. Is that true?
3: We currently make the corpora for 10 of the 17 fuzzers. I will get the corpora added for the new fuzzers over the next days.

PR #517, opened 2020-05-12
[pageseg.c] Added fuzzer

AdamKorcz:
This PR adds a fuzzer for targets in `pageseg.c`


DanBloomberg:
This looks OK.  It's a bit of a nit, but I'd prefer a uniform and simpler naming convention, as you've seen, with, e.g., pix1, pix2, ...., pixa1, pixa2, ....   box1, box2, ...  etc.

PR #518, opened 2020-05-12
[pix1.c] Added fuzzer

AdamKorcz:
This PR adds a fuzzer for targets in pix1.c that need to be explicitly called to be covered.


DanBloomberg:
you need to free the extracted data.  See pix1.c, l. 1941ff

lept_free(data2);

PR #519, opened 2020-05-14
Add allocation interception mechanism for embedded builds.

robinwatts:
Leptonica contains a setPixMemoryHandler mechanism for
intercepting allocations used for the data sections of
Pix (except on Windows where it is broken/disabled).
There is no mechanism for intercepting other such
allocations though.

Here, for the sake of embedded builds that need/want to
be able to control/track all allocations, we add such a
mechanism.

If LEPTONICA_INTERCEPT_MALLOC is defined during startup,
then all calls to malloc/calloc/realloc/free are redirected
to leptonica_{malloc/calloc/realloc/free} functions that
the library assumes will be provided for it.


DanBloomberg:
This seems OK.  Give me a couple of days to think about it & check it out.

DanBloomberg:
Thanks.  Works fine.

PR #530, opened 2020-08-19
Added 3 new fuzzers

AdamKorcz:
This PR adds 3 fuzzers.

Each file contains targets of the related source-file that must be called explicitly in order to be covered.

The targets in boxfunc3_fuzzer.cc may not seem obvious, and comments above each section clarify what the target of the given section is.

### DIFF
--- a/prog/fuzzing/boxfunc3_fuzzer.cc
+++ b/prog/fuzzing/boxfunc3_fuzzer.cc
@@ -1,0 +1,109 @@
+#include "leptfuzz.h"
+
+extern "C" int
+LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) { 
+	if(size<3) return 0;
+ 
+	leptSetStdNullHandler();
+
+	BOXA *boxa_payload, *boxa1;
+	boxa_payload = boxaReadMem(data, size);
+	if(boxa_payload == NULL) return 0;
+
+
+	PIX          *pixc, *pixd, *pix, *pixs;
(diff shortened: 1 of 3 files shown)


DanBloomberg:
I made a few trivial changes to two of the files.
If they build OK I'll merge it.

Perhaps hold off on more fuzzers for now, because I'm still trying to build the fuzzers using a shell script.  It has been surprisingly difficult.

Have you actually built these fuzzers, so that you could run them in a shell with a given payload?

AdamKorcz:
> Have you actually built these fuzzers, so that you could run them in a shell with a given payload?

Yes, these fuzzers run on OSS-fuzz's platform. I test all fuzzers on their infrastructure.

Regarding running the fuzzers: @stweil made a great contribution for this purpose. I haven't done it myself, but instructions to run a fuzzer locally can be found [here](https://github.com/DanBloomberg/leptonica/tree/master/prog/fuzzing#leptonica-fuzzing).

If that is not what you are looking to do, please feel free to share and I will see if there is something I can do.

I will check the build of the fuzzers and confirm.

DanBloomberg:
@stweil 
re: building fuzzer locally
What is the environment in which you run configure?  Do you first apt-get install clang?

stweil:
Yes, clang provides the best support for fuzzing, but newer versions of gcc also work at least partially (for certain sanitizers).

See [fuzzing documentation for Tesseract](https://tesseract-ocr.github.io/tessdoc/Compiling-%E2%80%93-GitInstallation.html#builds-for-fuzzing) for some hints.

PR #532, opened 2020-08-20
Fixed 2 missing semicolons

AdamKorcz:


DanBloomberg:
I found these when building the fuzzers, so they're already fixed.

PR #533, opened 2020-08-25
affine_reg: Fix memory leak

stweil:


stweil:
I found this leak while testing code with sanitizers built for fuzzing.

DanBloomberg:
thank you!

PR #538, opened 2020-09-22
CMake: update CMAKE_REQUIRED_INCLUDES when openjp2 library is found

BorisMansencal:
Currently openjpeg may be found by lines 100-102 in top CMakeLists, but the openjpeg include path will not be searched by check_include_files() in Configure.cmake.

I have a compilation error in a travis osx image. openjpeg headers are installed in /usr/local/include/ and found by the top CMakeLists.txt (and  so HAVE_LIBJP2K is defined). But the headers are not found by the Configure.cmake check_include_files(openjpeg-2.x/openjpeg.h ...) and so LIBJP2K_HEADER is not defined and the compilation of jp2kio.c fails because openjpeg.h is not found for the #include <openjpeg.h> line 117.

Adding ${JP2K_INCLUDE_DIR} to CMAKE_REQUIRED_INCLUDES, which are the paths searched by check_include_files() solves the problem.


DanBloomberg:
I'm going to leave this to the experts.

Note that the openjpeg library (jp2k) is NOT required for building leptonica.  (In fact, we don't require any image compression libraries, although most users will have at least tiff,png and jpeg.)

Does putting jp2k explicitly in CMAKE_REQUIRED_INCLUDES mean that it would be required?

Dan

BorisMansencal:
jp2k include directory is added to CMAKE_REQUIRED_INCLUDES only if jp2k was previously found by the CMakeLists.txt.
So this patch does not add a new requirement on jp2k and it is still possible to build leptonica without jp2k or any other image compression library.

DanBloomberg:
Good to know -- thank you for the explanation.
Will still leave for the experts to review.

PR #539, opened 2020-09-22
Avoid warning when several version of openjpeg are installed

BorisMansencal:
Currently, when several versions of openjpeg are found, LIBJP2K_HEADER is defined several times by cmake/Configure.cmake. gcc produces a warning `warning: "LIBJP2K_HEADER" redefined` on <build_dir>/src/config_auto.h

This patch ensures that LIBJP2K_HEADER is defined only once even when several versions of openjpeg are found.
It gives priority to the most recent version of openjpeg (2.3 > 2.2 > 2.1 > 2.0).


DanBloomberg:
Thank you -- much appreciated!  Those warnings are irritating!

PR #540, opened 2020-09-23
Revert "Avoid warning when several version of openjpeg are installed"

DanBloomberg:
Reverts DanBloomberg/leptonica#539


DanBloomberg:
The windows build failed, but I hadn't noticed that when I approved the merge.
So I've reverted the change.

The failure is running out of disk space, because it apparently keeps building progs multiple times !!
We can reconsider this PR when we understand the cause of the build failure.

DanBloomberg:
Not reverting.  This is not causing the problems with the windows buiild.

PR #246, opened 2017-04-20
cmake install configuration

aleksey-nikolaev:
I don`t know what exactly programs need to be installed, so I took the option from Makefile.am.
Probably fix #117

PR #544, opened 2020-10-10
Use correct location for CMake files

Shatur95:
CMake files should be located in `${CMAKE_INSTALL_PREFIX}/lib/cmake` instead of `${CMAKE_INSTALL_PREFIX}/cmake`, see [search procedure](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure). I fixed paths in this PR.


Shatur95:
Noticed another issue not related to this request. When using GNU Autotools, you copy `LeptonicaConfig.cmake`, which includes `LeptonicaTargets.cmake`, which is only generated with CMake. Therefore, if the project was installed using Autotools, then when you search for Leptonica in a CMake project, you will get an error because `LeptonicaTargets.cmake` file does not exist. Are you planning to support both build systems? If yes, is there a way to generate `LeptonicaTargets.cmake` with Autotools? Most Linux distributions use Autotools to build Leptonica.

zdenop:
... and a lot of project are switching to cmake. IMO generating `LeptonicaTargets.cmake` also with autotools make sense, so users are not push to use several building system for one project.

Shatur95:
I would like to help with this, but I am not familiar with GNU Autotools :(

However, this PR can be merged as is because it solves the CMake problem.

Shatur95:
@DanBloomberg, could this PR be merged?

DanBloomberg:
@egorpugin 

Please review. He's putting lib/ into the destination path.

egorpugin:
I'm not familiar with cmake install.
Those lines were contributed by @aleksey-nikolaev in https://github.com/DanBloomberg/leptonica/pull/246

Looking at the initial PR message `search procedure` (https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure) link, proposed paths are used for Unix while Windows uses another convention.
So, this PR might fix unix and break windows systems.

Shatur95:
>  So, this PR might fix unix and break windows systems.

The documentation says that `This is merely a convention, so all (W) and (U) directories are still searched on all platforms.`
So, Windows will work perfectly with `lib/cmake`. And this location is also common for Linux.
Previously, modules were installed directly into the `/cmake` folder, which is bad for Linux. At the moment all distributions just use autotools to install leptonica.

egorpugin:
> At the moment all distributions just use autotools to install leptonica.

Because cmake build was initially windows only.

> The documentation says that This is merely a convention, so all (W) and (U) directories are still searched on all platforms.

I'm merging the PR then.

Shatur95:
Thanks!

PR #545, opened 2020-10-10
Added 10 new fuzzers

AdamKorcz:


DanBloomberg (commented):
missing pixDestroy() on pix_copy1, pix_copy2, pix_copy3

AdamKorcz:
destroying pix_copy1, pix_copy2, pix_copy3 in morph_fuzzer gives a use-after-free when calling pixDestroy. Tested 20 minutes ago.

DanBloomberg:
You're right.  The interface to those functions is a bit tricky.   With the generality comes some danger.

Look at the function header.  If both the dest and src are defined, so not reassign the dest to another name, because then the original dest handle is pointing to the same pix data!

So call this way:
    pixCloseGeneralized(pix_copy1, pix, sel);
   pixDestroy(&pix_copy1);

and the same for the other two.

DanBloomberg:
Thank you!!!

PR #551, opened 2020-11-04
Removed assignment in adaptmap_fuzzer

AdamKorcz:
This PR removes the assignment of `pixContrastNorm` since the first parameter is not `NULL`


DanBloomberg (commented):
Note that in the input variables, either pixd == pixs or pixd == NULL.
Using the first, get rid of pix1 and call this way:
```
pixContrastNorm(payload_copy, payload_copy, 10, 10, 3, 0, 0);
```

DanBloomberg:
you don't need pix1, but let's leave it in, and I'll remove it on my next push

PR #554, opened 2020-11-26
Fix CID 1499794 (Dereference after null check)

stweil:


stweil:
The latest Coverity Scan uses the latest scan software version which reports false positives, but also correct positives like the one fixed here. The previous scan software version is no longer accepted by Coverity Scan.

DanBloomberg:
Thank you.  I believe I recently made that copy/paste error.

PR #556, opened 2020-12-05
Don't free data which was passed to functions with copyflag L_INSERT

stweil:
This fixes new duplicate free and use after free errors reported by
Coverity Scan.


DanBloomberg:
Thank you.  I missed that condition with a recent change!

PR #560, opened 2020-12-28
Added 4 fuzzers and rewrote existing fuzzer

AdamKorcz:
This PR adds 4 new fuzzers and a rewritten version of the enhance-fuzzer.

The reason for rewriting the enhance-fuzzer is to optimize the targets in the fuzzer. Those targets that are called elsewhere are removed and new ones have been added.

The enhance-fuzzer does leak memory. To me it looks like everything is cleaned up the way it should be, so any input in this regards is much appreciated.


DanBloomberg:
Looks good.  A couple of things:

* The leak in the enhance fuzzer is in l. 123: pixTRCMapGeneral() returns a pix.

* You don't need to check a pix before pixDestroy().  So in the enchance fuzzer, just do
  pixDestroy(&return_pix);

There are a few format issues that don't affect performance and I'll clean them up later.

AdamKorcz:
Thank you for the comments. They should have been addressed now.

DanBloomberg:
thank you!!

PR #563, opened 2021-01-18
Added more fuzzers

AdamKorcz:
This PR adds more targets to existing fuzzers as well as a new ccthin_fuzzer

### DIFF
--- a/prog/fuzzing/blend_fuzzer.cc
+++ b/prog/fuzzing/blend_fuzzer.cc
@@ -2,85 +2,92 @@
 
 extern "C" int
 LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) { 
-	if(size<3) return 0;
+    if(size<3) return 0;
 
-	leptSetStdNullHandler();
+    leptSetStdNullHandler();
 
-	PIX *pixs_payload = pixReadMemSpix(data, size);
-	if(pixs_payload == NULL) return 0;
+    PIX *pixs_payload = pixReadMemSpix(data, size);
+    if(pixs_payload == NULL) return 0;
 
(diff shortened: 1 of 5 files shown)


DanBloomberg:
In blend_fuzzer, you have a loop over "size".  But what if "size" is a large number?  Shouldn't it be clipped, or maybe just use a constant like 40 and use 7 * i instead of i in the parameters (to get over 255).

Otherwise looks fine.  There are a few unnecessary checks before deleting stuff (pix, sel).

DanBloomberg:
@egorpugin , @stweil 

travis failure due to being unable to find l_openjp2 at link time.

In the early part of the build, it failed to find 2.0, ... 2.3, so why was it in the list of libraries to link?
(Or is the list of libraries fixed, independent of the search?)

CMakeLists.txt looks for openjpeg-2.3.
I believe that 2.4 was recently released.  Is 2.3 no longer available in this build?

AdamKorcz:
> In blend_fuzzer, you have a loop over "size". But what if "size" is a large number? Shouldn't it be clipped, or maybe just use a constant like 40 and use 7 * i instead of i in the parameters (to get over 255).

Yeah, we shouldn't use size as the cutoff for loops. I tried it locally for a limited time and it was fine, but for continuous fuzzing we should avoid it. I will change that.

DanBloomberg:
@AdamKorcz 
Has this been updated?  I can't tell looking at 6eb1989

AdamKorcz:
I changed the loop to just keep it below 10 for now. I will consider another number one this has run and I can check how it runs continuously.

DanBloomberg:
Thank you!

PR #569, opened 2021-03-15
fix compilation with static zlib

neheb:
zlibVersion() is not in the static library.

### DIFF
--- a/src/libversions.c
+++ b/src/libversions.c
@@ -172,7 +172,7 @@ char    *versionStrP = NULL;
     if (!first) stringJoinIP(&versionStrP, " : ");
     first = FALSE;
     stringJoinIP(&versionStrP, "zlib ");
-    stringJoinIP(&versionStrP, zlibVersion());
+    stringJoinIP(&versionStrP, ZLIB_VERSION);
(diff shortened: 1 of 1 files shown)


DanBloomberg:
I'm using zlib.h version 1.2.11, from 2017.

It has both ZLIB_VERSION and zlibVersion().
For the latter, it says zlibVersion() is included for compatibility with versions < 1.0.2.

What version of zlib.h are you using?

neheb:
1.2.11

DanBloomberg:
So you're using a different library that doesn't have zlibVersion(), even though it is declared in zlib.h?

neheb:
I'm using leptonica...

neheb:
As I said, zlibVersion() is not in libz.a

DanBloomberg:
nm  -D /lib/x86_64-linux-gnu/libz.so.1

...
0000000000010050 T zlibVersion

However,
nm -D /usr/lib/x86_64-linux-gnu/libz.a

doesn't have it, as you say.

I see that 1.0.2 is from 1996.  That seems to be too old to worry about  :-)

PR #571, opened 2021-03-29
Remove old code which is unused since commit 2594c8b6 (fixes CID 1503…

stweil:
…048)


stweil:
@DanBloomberg, Coverity scan reports one more which is not a false positive:

    *** CID 1503047:  Resource leaks  (RESOURCE_LEAK)
    /src/pdfio2.c: 869 in l_generateJpegData()

DanBloomberg:
thanks for fixing this!

PR #573, opened 2021-04-11
Fix OSS-Fuzz issue 28181 (Use-of-uninitialized-value)

stweil:
The written data was not terminated with '\0' if HAVE_FMEMOPEN was defined,
so using strlen on it caused a buffer overrun.

### DIFF
--- a/src/boxbasic.c
+++ b/src/boxbasic.c
@@ -2094,6 +2094,9 @@ FILE    *fp;
     if ((fp = open_memstream((char **)pdata, psize)) == NULL)
         return ERROR_INT("stream not opened", procName, 1);
     ret = boxaaWriteStream(fp, baa);
+    fputc('\0', fp);
+    fclose(fp);
+    *psize = *psize - 1;
 #else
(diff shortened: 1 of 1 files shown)


stweil:
There are 20 other code locations which also use 0-terminated data for Windows, but not for Linux. I can send a pull request for those, too, if desired.

DanBloomberg:
I just saw this PR (have been away for the weekend).
Let me do some experiments, because I am surprised.
We haven't had this problem when writing to and reading from an actual file stream,
so why are we seeing it with a memstream?
Also, shouldn't the fix here be in boxaWriteStream() instead of boxaaWriteStream()?

stweil:
The fixes here are in `boxaWriteMem` and `boxaaWriteMem` because those functions write data to memory, so that data should be 0-terminated to allow using `strlen`. The OSS-Fuzz test case complained because `strlen` read undefined bytes after the valid data.

A terminating `\0` is not needed and usually also not desired when writing to files.

DanBloomberg:
Agree.  Thank you for doing all this !!

Issue #580, opened 2021-05-24
Bug reading PNM from stdin


giuliolunati:
# HOW TO REPRODUCE:

     #Create test.pnm
     cat << EOF > test.pnm
     P5
     12  1
     255
     123456789012
     EOF

     #Create test.c
     cat << EOF > test.c
     #include <leptonica/allheaders.h>
     int main(int argc, char **args) {
          PIX *p= pixReadStreamPnm(stdin);
          pixPrintStreamInfo(stderr, p, NULL);
     }
     EOF

     #Compile test:
     gcc -o test test.c -llept -ljpeg -lpng -ltiff

     #Read test.pnm from stdin
     cat test.pnm | ./test

     #The output says width=2 instead of width=12

# DIAGNOSIS:
The 1st char "1" in "12" is eaten away in `pnmSkipCommentLines()`, because `fseek(fp, -1L, SEEK_CUR)` can't work with `stdin`

giuliolunati:
Should be fixed by #581

DanBloomberg:
Fixed by #581

PR #581, opened 2021-05-26
pnmio.c: avoid fseek()

giuliolunati:
The goal is to make it possible
to read a PNM/PAM image from stdin (netpbm-like),
which is currently not working: see issue #580.

The reason is the use of fseek(fp, -1, SEEK_CUR)
to "go back" after a fgetc(),
which doesn't work for stdin,
so it has been replaced through fscanf().

### DIFF
--- a/src/pnmio.c
+++ b/src/pnmio.c
@@ -1352,13 +1352,12 @@ l_int32   c, ignore;
     *pval = 0;
     if (!fp)
         return ERROR_INT("stream not open", procName, 1);
-    do {  /* skip whitespace and non-numeric characters */
-        if ((c = fgetc(fp)) == EOF)
-            return 1;
-    } while (!isdigit(c));
 
-    fseek(fp, -1L, SEEK_CUR);        /* back up one byte */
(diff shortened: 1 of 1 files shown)


giuliolunati:
This could close #580

DanBloomberg:
Very elegant solution to the problem!
I am letting it build..
Verified it works on your test file using your patch.
Will do some other checks tomorrow and then merge.
Thank you very much for this.

Dan

giuliolunati:
And thank you very much to you for Leptonica!
I learned a lot from your code.
Now I'm working on a CLI utility for fixing my scanned docs.

DanBloomberg:
Is your name Giulio Lunati, by any chance  ;-)
I want to acknowledge you for your improvements to pnm reading.

Dan

giuliolunati:
Yes Dan, I'm Giulio Lunati (Pavia, Italy)
Thank you :-)

DanBloomberg:
Thank you again.
I will put some comments in line with the code.

PR #583, opened 2021-06-27
makeTempDirname: use confstr on macOS, and fix a bug in path generation

rsesek:
makeTempDirname() on iOS uses confstr() to get a temporary directory
because /tmp is not available to apps. macOS programs using the App
Sandbox are similarly restricted, and confstr is the recommended system
API to acquire a temporary directory path regardless.

This also fixes a bug introduced in
d15f1c24a16a65a4ae2e018a21ce9ad52b33ac64 where copying `path` to
`result` was changed from using strncpy() to stringCat(). Since the call
to confstr() already wrote to `result`, this led to the path being
incorrectly duplicated. Using stringCopy() resolves the issue.

### DIFF
--- a/src/utils2.c
+++ b/src/utils2.c
@@ -200,7 +200,7 @@
 #include <sys/types.h>
 #endif
 
-#ifdef OS_IOS
+#ifdef __APPLE__
 #include <unistd.h>
 #include <errno.h>
 #endif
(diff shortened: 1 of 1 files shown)


DanBloomberg:
Thank you for this PR!

This should be fine.  There's always a question if it will work properly with android, but we'll do those tests after the merge.

amitdo:
>- on windows, mac: by GetTempPath()
>- on ios: by confstr() (see man page)

This comment should be updated to reflect the change in the code.

DanBloomberg:
The obvious change is:
``` 
 *             - on Windows: by GetTempPath()
 *             - on macOS, iOS: by confstr() (see man page)
```
We never explicitly call out android, so it is handled with "/tmp" like unix.

DanBloomberg:
fixed

amitdo:
>fixed

In commit ffca513232f5fccf

PR #595, opened 2021-10-29
Fuzzing: Add CIFuzz

AdamKorcz:
Adds CIFuzz. This will run Leptonicas fuzzers for 600 seconds in the CI when a pull request is made. The time can be changed.

Documentation of CIFuzz can be found here: https://google.github.io/oss-fuzz/getting-started/continuous-integration/


DanBloomberg:
Thank you for this -- I had no idea this was offered by the oss-fuzz folks.

PR #599, opened 2021-12-04
CMake fixes

diizzyy:
Make lib(WebP) and OpenJPEG support optional
Report correct version in generated .pc file
Populate Libs.private in .pc file
Properly define library version
Install imagetops for consistency with Autotools build


diizzyy:
You're free to change whether these should be enabled or disabled by default

zdenop:
What is the rationale behind this? Why do you turn them off by default?

diizzyy:
As I said, feel free to use whatever option as you set fit as default. The main point is to make sure they aren't pulled in unintentionally.

diizzyy:
Remaning issues:
Library symlinks are a bit weird/inconsistent to autotools

```
lrwxr-xr-x  1 root  wheel       21 Dec  4 15:44 libleptonica.so -> libleptonica.so.5.4.0
-rwxr-xr-x  1 root  wheel  2849560 Dec  4 15:44 libleptonica.so.1.82.0
lrwxr-xr-x  1 root  wheel       22 Dec  4 15:44 libleptonica.so.5.4.0 -> libleptonica.so.1.82.0
```

DanBloomberg:
@egorpugin @stweil @zdenop
There are 5 relatively minor changes to cmake in this PR.
Do you have any issues with them?

Two questions:
(1) I don't know anything about private libs in cmake.  Are they useful?  Suggest that if they are adopted, the default be ON for jp2k and webp.
(2) Is there any reason for the use of the version number (e.g., 1.82.0) instead of the so number in the name of the library?

diizzyy:
Private.libs matters because if you have a downstream user that looks for the pc file (pkgconfig) it'll break.
As for so number this should be consistent with autotools

CMakeLists.txt
```diff
@@ -55,70 +55,93 @@ else()
...
+    option(OPENJPEG_SUPPORT "Build with OpenJPEG support" OFF)
 endif()
 
 set(leptonica_INSTALL_CMAKE_DIR "lib/cmake/leptonica"
     CACHE STRING "Install destination for CMake package files")
 
 if(NOT SW_BUILD)
     find_package(GIF)
+        if(GIF_FOUND)
+            set(libs_private "-lgif")
```
stweil, line 68:
```suggestion
            set(libs_private "${libs_private} -lgif")
```

CMakeLists.txt
```diff
@@ -55,70 +55,93 @@ else()
...
+    option(LIBWEBP_SUPPORT "Build with WebP support" OFF)
+    option(OPENJPEG_SUPPORT "Build with OpenJPEG support" OFF)
 endif()
 
 set(leptonica_INSTALL_CMAKE_DIR "lib/cmake/leptonica"
     CACHE STRING "Install destination for CMake package files")
 
 if(NOT SW_BUILD)
     find_package(GIF)
+        if(GIF_FOUND)
```
stweil, line 67: Why are this and the other `if` blocks not on the same indentation level as `find_package`?

CMakeLists.txt
```diff
@@ -55,70 +55,93 @@ else()
 endif()
 if(UNIX)
     option(SYM_LINK "Create symlink leptonica -> lept on UNIX" OFF)
+    option(LIBWEBP_SUPPORT "Build with WebP support" OFF)
+    option(OPENJPEG_SUPPORT "Build with OpenJPEG support" OFF)
```
stweil, line 59:
```suggestion
    option(LIBWEBP_SUPPORT "Build with WebP support" ON)
    option(OPENJPEG_SUPPORT "Build with OpenJPEG support" ON)
```

stweil (changes requested):

CMakeLists.txt
diizzyy: Is that correct since it's initializing the variable?

CMakeLists.txt
diizzyy: Good catch, I'll correct that

CMakeLists.txt
stweil: It isn't, unless `GIF_FOUND` was true. Any of the following `set` statements could initialize it if all preceding `if` blocks are false.

stweil: You could initialize `libs_private` before the first `if` block to an empty string if necessary. The resulting value of `libs_private` can have a leading space. I don't know whether that is a problem.

diizzyy:
@stweil 
Fixed, thanks for reviewing!

diizzyy:
Anything else that needs to be done and/or adjusted?

diizzyy:
Friendly ping

DanBloomberg:
Daniel, approving your PR is above my pay grade.  The only build system I maintain is the hand-tweaked makefiles.
My impression of the changes:
Make lib(WebP) and OpenJPEG support optional:       ok with default ON
Report correct version in generated .pc file:                  seems reasonable
Populate Libs.private in .pc file:                                     seems reasonable
Properly define library version:                                      good to follow standards
Install imagetops for consistency with Autotools build:  ok

Anyone want to approve or suggest changes?

stweil:
> I don't know anything about private libs in cmake. Are they useful?

`lept.pc` on Debian has `Libs.private: -lz -lpng16 -lz -ljpeg -lopenjp2 -lgif -ltiff -lwebp -lwebpmux -lwebp` (note that `-lwebp` is unnecessarily duplicated), so adding the same libraries in cmake builds looks correct.

I'm afraid I cannot say much to the remaining changes as I don't think that I am an expert for cmake. @egorpugin and @zdenop contributed most to `CMakeLists.txt`

DanBloomberg:
@diizzyy 

Please look at
   https://github.com/tesseract-ocr/tesseract/issues/3738

While trying to resolve that issue, we're brought back to the various library naming issues.

PR #611, opened 2022-04-16
Remove Coverity scan build status (too many false positives)

stweil:


DanBloomberg:
Thank you again for supporting this service for five or six years, nearly the whole time leptonica has been on github.

Dan

PR #612, opened 2022-04-22
Fix documentation

stweil:


DanBloomberg:
Thanks -- I do prefer Refcount to RefCount.  Not sure why, but maybe because it indicates a bigger (more German?) type of noun.  :-)

PR #613, opened 2022-04-23
Fix some issues which were reported by Coverity Scan

stweil:


stweil:
I just started a new run with Coverity Scan which hopefully eliminates most reports which were related to reference counts.

DanBloomberg:
Thanks.  All worth doing!

PR #616, opened 2022-04-25
Remove PROCNAME / procName and use __func__

stweil:

### DIFF
--- a/src/rbtree.c
+++ b/src/rbtree.c
@@ -136,11 +136,9 @@ l_rbtreeCreate(l_int32  keytype)
 {
 L_RBTREE  *t;
 
-    PROCNAME("l_rbtreeCreate");
-
     if (keytype != L_INT_TYPE && keytype != L_UINT_TYPE &&
         keytype != L_FLOAT_TYPE && keytype)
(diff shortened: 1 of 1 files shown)


stweil:
This example shows how `PROCNAME` and `procName` can be removed. It could be applied to the whole code. That would avoid wrong function names, reduce the lines of code and avoid compiler warnings for functions which declare `procName` without using it.

DanBloomberg:
I didn't know about __func__.
Do you know how long that has been in effect?
Big job to apply to 2700+ functions  !
I could try a few files next time I feel like listening to a podcast  :-)

stweil:
Maybe I can help. Most changes are trivial, only some templates need special care. I now tried to update `prog` in pull request #617.

stweil:
> Do you know how long that has been in effect?

`__func__` was added by the C99 standard (see https://gcc.gnu.org/onlinedocs/gcc/Function-Names.html), more than 20 years ago. Luckily it is also supported by Microsoft compilers (at least since Visual Studio 2015, see https://docs.microsoft.com/de-de/cpp/preprocessor/predefined-macros?view=msvc-140).

PR #617, opened 2022-04-25
Remove PROCNAME / procName and use __func__

stweil:

### DIFF
--- a/prog/adaptmap_dark.c
+++ b/prog/adaptmap_dark.c
@@ -70,8 +70,6 @@ L_BMF   *bmf;
 PIXA    *pixa1, *pixa2, *pixa3, *pixa4;
 PIXAC   *pixac1, *pixac2, *pixac3;
 
-    PROCNAME("adaptmap_dark");
-
     setLeptDebugOK(1);
     bmf = bmfCreate(NULL, 10);
     index = 0;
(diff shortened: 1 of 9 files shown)


DanBloomberg:
Some of these deal with automatic generation of code, so it's a bit more complicated.
You should leave the following files out of this PR, because they are autogenerated
autogen.137.c
dwalinear.3.c
(fmorphgen.1.c and dwacomb.2.c are also autogen'd)

Also best to leave out the changes to the three *template1.txt files.
When these template files are modified, the programs that call them must also be modified.
For example, morphtemplate1.txt is used by fmorphauto.c, which generates the code to make the PROCNAME("...") calls.
That code also needs to be removed; e.g., the strings in l. 348-341 in fmorphauto, as well as the use of those strings in l. 422 and l. 438.
After the template and fmorphauto are changed and recompiled, you must run dwalineargen to generate dwalinear.3.c and dwalinearlow.3.c.
The fmorphauto functions are also used by fmorphautogen to make fmorphgen.1.c,
and by fcombautogen to make dwacomb.2.c.
Finally, all the newly generated code needs to be compiled and the tests run.  So you should probably let me worry about the autogen'd stuff  :-)

If you spend a while around this stuff -- which I don't recommend, by the way -- you might notice that fmorphgen*.1.c and dwacomb*.2.c have been put in src, but dwalinear*.3.c are in prog where they were generated.  The reason is that the low-level code in dwalinearlow.3.c is so big that I didn't want to include it in the library by default.

Same goes for hmttemplate1.c, which along with functions in fhmtauto (after recompiling) are then used by the program fhmtautogen to generate fhmtgen*.1.c
And stringtemplate1.c requires multiple steps as well, though it's not quite as complicated.

stweil:
I now removed these files from the commit:

    prog/autogen.137.c
    prog/dwalinear.3.c
    prog/hmttemplate1.txt
    prog/morphtemplate1.txt
    prog/stringtemplate1.txt

PR #635, opened 2022-08-08
Fix oss-fuzz build failure

DavidKorczynski:
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=49777
Fixes: https://github.com/google/oss-fuzz/issues/8152

### DIFF
--- a/prog/fuzzing/binarize_fuzzer.cc
+++ b/prog/fuzzing/binarize_fuzzer.cc
@@ -45,10 +45,13 @@ LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
 	pixDestroy(&pix_copy1);
 
 	pix_copy1 = pixCopy(NULL, pixs_payload);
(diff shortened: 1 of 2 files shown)


DavidKorczynski:
CC @AdamKorcz

stweil (approved):
Thank you for fixing this oss-fuzz build issue.

DavidKorczynski:
The CIFuzz failure should be due to a limitation in CIFuzz rather than this fix. When I substitute https://github.com/google/oss-fuzz/blob/c1bcbc2c54d0d28de5c0556ad1a0b817ea0ae291/projects/leptonica/Dockerfile#L20 for `RUN git clone https://github.com/DavidKorczynski/leptonica && cd leptonica && git checkout leptonica-ossfix` and run the OSS-Fuzz `check_build` then I get `Check build passed.` and the build succeeds.

DanBloomberg:
Thank you David for maintaining the leptonica fuzz scripts and fixing this.  I'll merge and we'll see if CIFuzz is fixed.

PR #636, opened 2022-08-08
Fix minor nit to actually fix OSS-Fuzz build

DavidKorczynski:
Fixes the issues from here
https://github.com/DanBloomberg/leptonica/pull/635

After https://github.com/DanBloomberg/leptonica/pull/635 was merged I ran a clean build from OSS-Fuzz and noticed there was indeed a mistake. Unfortunately I had not cleared my cache in the $WORK dir previously, which meant the mistake did not show up locally. This should fix it and I believe CIFuzz will also pass this time.

### DIFF
--- a/prog/fuzzing/oss-fuzz-build.sh
+++ b/prog/fuzzing/oss-fuzz-build.sh
@@ -55,7 +55,8 @@ CFLAGS="$WEBP_CFLAGS" ./configure \
   --disable-jpeg \
   --disable-tiff \
   --disable-gif \
-  --disable-wic
+  --disable-wic \
+  --prefix=$WORK
 make clean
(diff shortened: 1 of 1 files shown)


DanBloomberg:
That was fast!  Thank you again!!

DavidKorczynski:
Thanks @DanBloomberg !

The fuzzers were built and run this time, but the CIFuzz job failed as one of the fuzzers hit a memory leak here: https://github.com/DanBloomberg/leptonica/runs/7729625274?check_suite_focus=true

Now the fuzzers will be built by OSS-Fuzz again though! The OSS-Fuzz bot should check this in the next 24 hours and close the Monorail issue.

PR #639, opened 2022-08-24
Add vcpkg installation instructions

FrankXie05:
`leptonica` is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `leptonica` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `leptonica`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/leptonica/portfile.cmake). We try to keep the library maintained as close as possible to the original library. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,3 +80,15 @@ www.leptonica.org
   * Zdenko Podobny: Zdenko has worked, mostly behind the scenes as a primary maintainer of tesseract, to help with leptonica builds on all platforms, and coordinate with its use in tesseract.
   * Adam Korczynski: Adam is an expert in testing libraries for safety.  He has built most of the open source fuzzers for leptonica in the oss-fuzz project, with significant code coverage.
 
+## Installing leptonica (vcpkg)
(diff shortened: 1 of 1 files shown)


DanBloomberg:
Impressively succinct script!
Did you find out why openjpeg (jp2k) failed to link in your build?

Thank you for the contribution!

Dan

FrankXie05:
@DanBloomberg Thank you very much for your quick response and acceptance.
Regarding openjpeg, can you provide any issue or pr? I will do my best to solve it. :)

DanBloomberg:
Both autoconf and cmake are able to link openjpeg to leptonica.
I'd suggest you look at those scripts for guidance.

One irritating thing about openjpeg, which may be the source of your troubles, is that, unlike all the other I/O libraries, it puts the header files in a subdirectory specific for the release.  This requires extra conditionals in the scripts.

PR #648, opened 2022-11-20
Fix typo callof -> calloc and fix six Coverity Scan issues

stweil:


stweil:
The leaks are not really significant after an out of memory error, but fixing them is easy and makes Coverity Scan happy.

DanBloomberg:
Thanks!

PR #649, opened 2022-11-20
Fix memory leak (Coverity Scan #1503047)

stweil:


stweil:
The leak would be triggered with illegal values of `ascii85flag` (for example `2`).
The new code handles such values like `1`. Would a check for valid values be better?

DanBloomberg:
I think a check at the top for valid values of ascii85flag is a better way to handle this.

stweil:
Updated, now checks `ascii85flag`.

PR #650, opened 2022-11-21
Remove LGTM badges (LGTM.com will be shut down in December 2022)

stweil:

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,7 @@
 
 [![Build Status](https://api.travis-ci.com/DanBloomberg/leptonica.svg?branch=master)](https://app.travis-ci.com/github/DanBloomberg/leptonica)
 [![Build status](https://ci.appveyor.com/api/projects/status/vsk607rr6n4j2tmk?svg=true)](https://ci.appveyor.com/project/DanBloomberg/leptonica)
-![Build status](https://github.com/DanBloomberg/leptonica/workflows/sw/badge.svg)<br>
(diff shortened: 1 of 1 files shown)


DanBloomberg:
Darn -- I liked them.

amitdo:
Dan,

https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/

The replacement is CodeQL. You're already have [a CodeQL workflow](https://github.com/DanBloomberg/leptonica/blob/master/.github/workflows/codeql-analysis.yml).

However, It seems you need to [change some setting in the repository to actually see the alerts](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/setting-up-code-scanning-for-a-repository#setting-up-code-scanning-using-starter-workflows).

Here is the how CodeQL's alerts are presented in Tesseract:

https://github.com/tesseract-ocr/tesseract/security/code-scanning

DanBloomberg:
I can't see the tesseract/security/code-scanning link.

What workflow we should use with codeQL seems above my pay grade.
Running the analyzer after every push seems like overkill, and I don't know the costs or the benefits.
If Stefan has an opinion about what makes sense, I'd be happy to have him handle it.

amitdo:
Maybe only the people that are members of the repo can see the alerts.

stweil:
Yes, although CodeQL seems to be made by the same people as LGTM it looks much more complicated (like a typical Microsoft product). I still did not find out how to enable all reports from LGTM in CodeQL. Currently it only shows severe errors.

It is possible to configure how often continuous integration actions should run. I also wonder whether some runs might be overkill, and we should also consider that they cost electrical energy of course.

The security alerts are only visible for certain team members and selected people. That's similar for Coverity Scan, too. So Amit can add Dan to Tesseract, and Dan can add Amit to Leptonica. :-)

PR #651, opened 2022-11-21
Fix memory leak in pageseg_reg

stweil:

### DIFF
--- a/prog/pageseg_reg.c
+++ b/prog/pageseg_reg.c
@@ -233,5 +233,6 @@ L_REGPARAMS  *rp;
     pixDestroy(&pix2);
     pixDestroy(&pix3);
     pixDestroy(&pix4);
+    pixaDestroy(&pixadb);
     return regTestCleanup(rp);
 }


DanBloomberg:
Thanks.  Was that found by coverity scan?

stweil:
No, only indirectly. I had run `pageseg_reg` to examine a leak which was reported by coverity scan and detected this one.

amitdo:
Dan, Stefan,

How about adding a workflow that uses clang sanitizers with all Leptonica's programs? You can set it to run daily or weekly.

DanBloomberg:
That might be too much for me to deal with, even assuming it is possible somehow to turn 300 programs into sanitizers.  One thing we learned about sanitizers is that you don't want to test existing I/O libraries with it (it finds stuff you can't deal with), so for functions that need an image (most do), you need to use a simple encoding scheme -- Adam Korczynski and I settled on spix (un-encoded raster image) for payloads.

stweil:
A first result of a local test with sanitizer flags shows indeed many issues (mainly from graphic libraries), but also detected the issue which I fixed in pull request #652.

stweil:
The address sanitizer deteced some more issues. I added fixes for those to #652 now.

amitdo:
Stefan,

Another thing you can run on Leptonica's codebase is clang-tidy (it also have all the default checkers included by scan-build).

PR #652, opened 2022-11-23
Fix rankhisto_reg test by fixing square root of negative number

stweil:
The test fails in a debug build with compiler flag -ftrapv. That compiler flag modifies the code to detect illegal integer calculations.

An integer overflow was triggered because the difference of two floating point values was negative (rounding error) and the square root of negative values is undefined (NAN).

Converting NAN to integer results in an unexpected integer value.

### DIFF
--- a/prog/iomisc_reg.c
+++ b/prog/iomisc_reg.c
@@ -254,6 +254,7 @@ L_REGPARAMS  *rp;
     regTestCompareValues(rp, tiffsize[5], size, 0.0);  /* 29 */
     if (rp->display)
         lept_stderr("lzw: %ld\n", (unsigned long)size);
+    pixDestroy(&pixs);
 
         /* Test read/write of alpha with pnm */
     pixs = pixRead("books_logo.png");
--- a/prog/numa3_reg.c
+++ b/prog/numa3_reg.c
@@ -45,7 +45,7 @@ int main(int    argc,
          char **argv)
 {
 char          buf1[64], buf2[64];
-l_int32       i, n, hw, thresh, same, ival;
+l_int32       i, hw, thresh, same, ival;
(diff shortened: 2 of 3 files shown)


stweil:
I added two more commits which fix memory issues detected by the address sanitizer when running `make check`.

DanBloomberg:
Thank you.  I have used valgrind on the programs, but not the clang sanitizers.

Please say more about the "instructions for reproduction".  We should have them written out so that anyone (like, me  :-) can run sanitizers.  What needs to happen to configure.ac?

stweil:
> Please say more about the "instructions for reproduction".

I'll try to create a GitHub action with sanitizers. Like that all steps are also documented and can be reproduced.

DanBloomberg:
Thank you.  I've added a section  to the building part of the makefile.
Also ran the sanitizer to find unused variables.

DanBloomberg:
@stweil 
When the address sanitizer is run, there are many warnings, which make for a very noisy output.

The warnings are about variables that may be uninitialized (they seem ok), return values that are set but not used (needed to silence compilers and analyzers), static buffers made by the compiler that are larger than 4096 bytes (not a problem), variables that are not used because they're in a section of code that is not compiled, etc.

Because of all the noise, it's not clear to me how useful it would be as a github action.
But I did 'document' your script in README.html

DanBloomberg:
If I configure without those two flags, and then run ```make clean``` I still get warnings about:
* redefinition of preprocessor variables
* variables defined but not used
* variables that may be used uninitialized
* static strings longer than 4095 bytes
* variables set but not used
* using format flat %p with any pointer other than void *

There are about 200 of these, essentially all "noise".

PR #666, opened 2023-01-30
CMake: Use PRIVATE since the deps are not used in the headers

autoantwort:


zdenop:
I think it is ok.

DanBloomberg:
Merged.

PR #667, opened 2023-01-30
Fix webp transitive dependencies

autoantwort:
Includes #664

Currently not all transitive dependencies are forwarded in the generated cmake config which results in build failures downstream. 
For example tesseract fails with 
```
[...]

PR #674, opened 2023-02-21
Improve pkgconfig file

autoantwort:
Extracted from #667 (42ea1f2)

One should use Requires.private if you link against another projects that support pkg-config https://people.freedesktop.org/~dbn/pkg-config-guide.html#using


zdenop:
thanks.

PR #680, opened 2023-03-20
couple of nullptr checks to prevent crashes (MSVC2022)

GerHobbelt:


DanBloomberg:
I've been fairly consistent checking mallocs that can be quite large, but not worrying about tiny mallocs of a few hundred bytes at most.  The check for cdir returned from stringNew() in l.3086  should be checked.

GerHobbelt:
👍 Thanks again!

PR #682, opened 2023-03-20
(feature) append filename to any error messages where this would be useful

GerHobbelt:
append filename to any error messages where this would be useful (and available), so the user can locate the culprit more easily.


DanBloomberg:
This is an interesting patch. I like framing it generally as a 2nd argument (the 1st being the problem, the 3rd the procedure name and the 4th the return value.

One question:  Does this modify all the instances in the library where an image file is being read?  

One observation: this does not need to be used in programs in the prog directory.  (You may have included those 2 instances for testing?)   For regression programs, we have sufficient information about failures when they occur.  And other programs are relatively short and if they fail, debugging can easily be done if necessary.

DanBloomberg:
Actually, I answered my own question:
grep fopenReadStream *.c | wc
     58     400    3418
grep fopenWriteStream *.c | wc
     45     347    2981

We have over 100 files and 700 instances where these new macros could be used for reading and writing files.
How did you pick the subset that you modified?

DanBloomberg:
Thank you very much for this improvement in the leptonica infrastructure.

PR #683, opened 2023-03-20
`pixDisplayWithTitle()`: supports 'open' mode on Win32/64 + supports 'store' mode for all

GerHobbelt:
`pixDisplayWithTitle()`:
- supports 'open' mode on Win32/64 (opens the system-registered image viewer application via `explorer /open,%1`
- additional modes for all platforms:
  + NONE  (disables display entirely)
  + STORE (stores the image to /tmp/lept/disp/ as is done in the other modes, but DOES NOT invoke any external application for viewing, i.e. no `system()` call.

### DIFF
--- a/src/pix.h
+++ b/src/pix.h
@@ -1036,11 +1036,13 @@ enum {
  *-------------------------------------------------------------------------*/
 /*! Display Program */
 enum {
-    L_DISPLAY_WITH_XZGV = 1,  /*!< Use xzgv with pixDisplay()              */
+	L_DISPLAY_WITH_NONE = 0,  /*!< Do not use any: disable pixDisplay()    */
+	L_DISPLAY_WITH_XZGV = 1,  /*!< Use xzgv with pixDisplay()              */
     L_DISPLAY_WITH_XLI = 2,   /*!< Use xli with pixDisplay()               */
     L_DISPLAY_WITH_XV = 3,    /*!< Use xv with pixDisplay()                */
-    L_DISPLAY_WITH_IV = 4,    /*!< Use irfvanview (win) with pixDisplay()  */
-    L_DISPLAY_WITH_OPEN = 5   /*!< Use open (apple) with pixDisplay()      */
+    L_DISPLAY_WITH_IV = 4,    /*!< Use irfvanview (ms windows) with pixDisplay()        */
+    L_DISPLAY_WITH_OPEN = 5,  /*!< Use open (apple / ms windows) with pixDisplay()      */
+	L_DISPLAY_WITH_STORE = 6  /*!< Save the image to disk (in /tmp/lep/disp/ )          */
 };
 
 /*-------------------------------------------------------------------------*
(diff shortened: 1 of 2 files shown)


DanBloomberg:
I don't see the point in most of this.

We have a global L_DISPLAY_WITH_... that can be set with l_chooseDisplayProg(),
but for most situations the default suffices.  If you don't want output, use
0 for the dispflag in pixDisplayWithTitle -- we don't need to call
l_chooseDisplayProg(L_DISPLAY_WITH_NONE) first.  If you want to write the
image to file, use pixWrite() -- there's no good way to do it from
pixDisplay(), and your code doesn't do it either.  There is no place to
specify the output filename.

Did you mean to have the default windows display program be "open"?
If so, why that and why not 

For windows, what is wrong with getting the path on the stack with:
 ```
    pathname = genPathname(tempname, NULL);
    _fullpath(fullpath, pathname, sizeof(fullpath));
 ```
?

For Windows, we can allow use of the 'open' display program with this
variant of your code:
```
        /* Windows: L_DISPLAY_WITH_IV or L_DISPLAY_WITH_OPEN */
    pathname = genPathname(tempname, NULL);
    _fullpath(fullpath, pathname, sizeof(fullpath));
    if (var_DISPLAY_PROG == L_DISPLAY_WITH_IV) {
        if (title) {
            snprintf(buffer, Bufsize,
                     "i_view32.exe \"%s\" /pos=(%d,%d) /title=\"%s\"",
                     fullpath, x, y, title);
        } else {
            snprintf(buffer, Bufsize, "i_view32.exe \"%s\" /pos=(%d,%d)",
                     fullpath, x, y);
        }
    } else {  /* L_DISPLAY_WITH_OPEN */
        snprintf(buffer, Bufsize, "explorer.exe /open \"%s\"", fulllpath);
    }
```

Interesting syntax with the comma in the command line:
        explorer.exe /open,\"%s\"

I'll add checks at the top of the function to make sure that the display program is valid for the platform.

GerHobbelt:
BTW:

> If you don't want output, use
> 0 for the dispflag in pixDisplayWithTitle

yep, I noticed, but that's not going to fly for pixDisplay() calls, such as
in the /prog/*.c files, because that one hardcodes display=1 as part of the
arguments it passes to pixdisplayWithTitle() under the hood.



Met vriendelijke groeten / Best regards,

Ger Hobbelt

--------------------------------------------------
web:    http://www.hobbelt.com/
        http://www.hebbut.net/
mail:   ***@***.***
mobile: +31-6-11 120 978
--------------------------------------------------

DanBloomberg:
Thank you for the information about how Windows calls display programs.

And you're right, it would be useful to be able to disable pixDisplay() calls!

DanBloomberg:
commit 0ecd513
Please let me know if is is now working properly for you.

Dan

GerHobbelt:
👍 merged and working fine.

P.S. I like my STORE mode (so I can run `/prog/*`, etc. in batch mode without viewers popping up into my face, but still collect the images they were supposed to display), so I merged & mixed it back into my own codebase. But that's just me; won't be all that useful to others.

GerHobbelt:
Shall I close this as completed, then?

Issue #675, opened 2023-03-08
RGB order changed when calling pixReadWithHint/pixWrite


sinall:
```
auto pix = pixReadWithHint("source.bmp", IFF_BMP);
auto ret = pixWrite("target.bmp", pix, IFF_BMP);
```

The `source.bmp` is captured by ATL CImage on Windows 11.

Testing images are attached.

![image](https://user-images.githubusercontent.com/1246382/223670543-b6600c40-a18c-42eb-89de-2ddc7443421b.png)
[images.zip](https://github.com/DanBloomberg/leptonica/files/10919028/images.zip)

DanBloomberg:
pixReadWithHint() ignores the second arg if the input image format is not jpeg.  So there's no issue there.

I wrote the following test code, run on linux, and there are no issues with component order.  Try it, and let me know the results on Windows.

```
    pix1 = pixRead("source.bmp");
//    pix1 = pixReadWithHint("source.bmp", IFF_BMP);   // ignores 2nd arg
    pixWrite("/tmp/junk1.bmp", pix1, IFF_BMP);
    pixWrite("/tmp/junk1.png", pix1, IFF_PNG);
    pix2 = pixRead("/tmp/junk1.bmp");
    pix3 = pixRead("/tmp/junk1.png");
    pixEqual(pix2, pix1, &same);
    lept_stderr("same: %d\n", same);
    pixEqual(pix3, pix1, &same);
    lept_stderr("same: %d\n", same);
    pixDisplay(pix2, 300, 100);
    pixDisplay(pix3, 500, 100);
    pixWrite("/tmp/junk2.bmp", pix2, IFF_BMP);
    writeImageFileInfo("/tmp/junk1.bmp", stderr, 0);
    writeImageFileInfo("/tmp/junk2.bmp", stderr, 0);
```

sinall:
> pixReadWithHint() ignores the second arg if the input image format is not jpeg. So there's no issue there.
> 
> I wrote the following test code, run on linux, and there are no issues with component order. Try it, and let me know the results on Windows.
> 
> ```
>     pix1 = pixRead("source.bmp");
> //    pix1 = pixReadWithHint("source.bmp", IFF_BMP);   // ignores 2nd arg
>     pixWrite("/tmp/junk1.bmp", pix1, IFF_BMP);
>     pixWrite("/tmp/junk1.png", pix1, IFF_PNG);
>     pix2 = pixRead("/tmp/junk1.bmp");
[...]

PR #685, opened 2023-03-21
24bpp PIX to BMP would fail - fixes #684

### LINKED ISSUE #684: Writing 24bpp (depth=24, spp=3) PIX to BMP produces b0rked files
See also [comment at #675](https://github.com/DanBloomberg/leptonica/issues/675#issuecomment-1477757027) for code extract as sample code zip.

Relevant code sequence extract shown here (knowing `pixReadWithHint` is same as `pixRead` or BMP, but this was coded while looking at #675, where `pixReadWithHint` was initially used) :

```
	pix[5] = pixReadWithHint(DEMOPATH("test-rgba.bmp"), IFF_BMP);
	// identical to: pix[5] = pixRead(DEMOPATH("test-rgba.bmp"));

	d = pixGetDepth(pix[5]);
	assert(d == 32);
	spp = pixGetSpp(pix[5]);
	assert(spp == 4);
	pix[11] = pixConvert32To24(pix[5]);  // <<-- required to reproduce this issue: only happens for 24bpp PIX
	d = pixGetDepth(pix[11]);
	assert(d == 24);
	spp = pixGetSpp(pix[11]);
	assert(spp == 3);
	ret |= pixWrite("/tmp/lept/bmp-test/target-rgba24.bmp", pix[11], IFF_BMP);
	ret |= pixWrite("/tmp/lept/bmp-test/target-rgba24.png", pix[11], IFF_PNG);
```

Note that the expected output BMP filesize is around 15K (4 planes source is 20K; we dropped the A plane, so 25% size reduction estimate), but instead is a wild, quite large size.

Turns out d=24 will drive the bmp writer to use a cmapped / b<=8 code path, rather than the d=32/d=16 'full color' code path.

## A pullreq with a patch is forthcoming.
(I need to extract/cherrypick that from my fork, so expect a ~30 mins delay on that.)

GerHobbelt:
fixes #684

😞 I now see indentation is a wreck (editor set to TABS here, sorry)

### DIFF
--- a/src/bmpio.c
+++ b/src/bmpio.c
@@ -505,6 +505,9 @@ RGBA_QUAD  *pquad;
         L_WARNING("2 bpp files can't be read; converting to 8 bpp\n", __func__);
         pix = pixConvert2To8(pixs, 0, 85, 170, 255, 1);
         d = 8;
+	} else if (d == 24) {
+        pix = pixConvert24To32(pixs);
+		d = 32;
     } else {
         pix = pixCopy(NULL, pixs);
     }


DanBloomberg:
Thanks!

Does this really fix the windows issue with the test code that I provided for 4 component bmp image read/write?  The puzzle for me is that it works on linux but gets the color components borked on windows.

Note that leptonica uses 4 space indent without tabs.

GerHobbelt:
> Does this really fix the windows issue with the test code that I provided for 4 component bmp image read/write? The puzzle for me is that it works on linux but gets the color components borked on windows.

Nope. (I take it you mean #675?)

- Re #675: AFAICT, OP's problem is **weird** 🤔 : analysis (on my Win10 dev box) leads me to believe the problem is *outside leptonica*; I tried several ways (which uncovered this issue as a side effect) but was utterly unable to reproduce OP's issue. The most curious thing there is that the bpp changed from 32 to 24 according to his source+target bmp files, which I can't see happen anywhere in those leptonica APIs.  So (IMO) the puzzle is not about MSwin, but about the precise build & run environment of #675 OP.

  *Apologies if my analysis report comment there is fuzzy/unclear.* 
  I dug through his image files and tried a few things, including trying to answer the question: "*as it doesn't reproduce as-is, would it reproduce if one or two leptonica calls in between were left out in the issue report?*" 
  Result: I wasn't able to reproduce his `target.bmp` no matter what I did. I got *close*, via adding a `Convert32To24` call, but then the BGR channels are still perfectly correct and his had swapped B<->R, which is *just weird* as it's not explicable by an additional missing Endian swap. And this was already *assuming gaps in the report, second-guessing missing intel from the original report*. The other thing that's still open is: is he using a prebuilt lib from somebody else?  Anyway, it's a WTH any which way I look at it.

- Re #686: that one popped up while I wrote test code for #675. #686 is still pending; I haven't spent time looking into it further as it's lower prio for me now -- I still need to analyze & write demo/test code to file a few thresholding related oddities that I observed while testing tesseract and I consider those issues more pressing.


## summary

- #675 is still pending and a conundrum. IMO: possibly *force majeure* until OP chimes in with more info, as it's currently not reproducible on Win64 OS. Not with his code. Not with your snippet. Nor with mine. 🤷 
- #686: pending until further investigation. I expect a boo-boo in pixConvert32To24() for odd image widths, but that can wait.

GerHobbelt:
Thanks for the compliment. :-)
I still only somewhat grok the periphery; the important stuff, where the
math becomes important, is a *real struggle* still -- not your code, mind
you, just me hitting my own math limitations (edu + grasp) pretty hard. :'-(

example: i don't get why the counts in the histograms are *multiplied* by
their index (= grayscale tone 0..255) when producing the sum. this is done
when looking for a threshold in there (double hump in histogram -> seek
valley in between). :-(  that multiply feels like a triangle of weighting
which I don't see the benefit or use for yet; meanwhile I'm scanning otsu
and related papers to see if somebody says something in there that makes my
brain +click+ but thus far: no luck. alas, maybe today. one can hope. :-)

---

re 684: I was fast on the trigger as I assumed you'ld pick up 685 (the fix)
from the noises you made. :-)

(and, yes, I gathered that 24bpp is an oddity -- 4 bytes per pixel makes a
lot of sense hardware wise, while 3 bytes per px is just a lot of old skool
byte shuffling -- but I found convertTo24 in all headers while hunting ways
to make 675 happen, and then it helps when write\io interfaces are
orthogonal (if that's the right word for this): accept anything, either
directly or by implicit conversion, only reject the obviously wrong.  Your
library behaviour is like that, which makes it relatively easy to go
through (few exceptions to keep in mind) and use. I just grabbed what I
found, than hit a snag, 's all.

re 675: yes. I cannot reproduce, nor reason through code (leptonica & his +
data images provided) why it might happen. sinall is the only one. alas.


(AFK, responding via mobi. apologies for degraded msg fmt)

DanBloomberg:
OK, at this time I can assume that the BMP code is working properly for 4 spp on all platforms.

Small amount of background on this.  I implemented the BMP I/O shims years ago because people needed it, but I did so with minimal features.  For example, I didn't even allow reading of RGBA.  My mantra was: if you want to use the transparency component, use PNG.  Then about 6 months ago people started complaining that it didn't handle RGBA that they were getting from screen grabs.  I still didn't want to mess with it, so just checked for 4 spp and returned with an error message.  Then someone sent some code (isn't github great!) that worked for reading a BMP RGBA file and successfully tossed out the A layer.  At that point I gave in, and implemented both reading and writing RGBA with BMP, including keeping the transparency.   (I included writing both for completeness and to check that the reading was correct.  As you saw in the code snippet (and you did this as well with your testing), the files were read and written several times, and also with write/read in PNG to make sure the images were all the same.

But I only tested on linux; hence the concern with issue #675.  Will leave this open in case sinall reports what he is seeing in detail (as requested).

As for Otsu thresholding, it's commonly used but not very useful because it is a global threshold.  I made some modifications to make it work a little better, but it's still very poor.   leptonica has adaptive thresholding for doing a decent job on grayscale images where there is a large variation in background intensity.  You can see examples of this by running:
      adaptnorm_reg display

DanBloomberg:
This PR is no longer relevant.

I want to thank Ger again for taking the issue seriously, and also for providing a very useful modification to leptonica error handling -- allowing a second string in the error message, most useful when there is a failure to read a file.

PR #711, opened 2023-08-24
Fixes for MSVC2022 warnings about various lines of code

GerHobbelt:
(as it says on the tin; see also the commit messages)

Mostly about "possible loss of significant digits" when numeric constants are to be treated as `float`, several places where the compiler yaks about "possible use of uninitialized variable/pointer" (which could be argued to be due to the analyzer to be slightly retarded/overzealous, as [PVS](https://pvs-studio.com/en/) wasn't complaining about most of these but I rather lean towards appeasing that compiler than missing other useful warnings from the same due to this noise cluttering the Output log when building my libraries.

No performance loss, all benign edits as far as I am concerned.


DanBloomberg:
these changes look reasonable.  Did you run the regression tests ```alltests_reg``` before/after making them?

DanBloomberg:
Never mind.  I patched and rand the tests.  
Four tests failed, all due to the small changes to projective.c, where a double division was replaced by a float division (apparently).  The changes in the output are completely insignificant, so it's fine with all your changes.

PR #712, opened 2023-08-24
minor Numa safety patches

GerHobbelt:
See also the commit messages... I hit upon these thanks to my own misuse of one these APIs (unintentionally feeding it a NUMA array of zero length) and I got some random weird results with MSVC2022, then I investigated and found `numaGetCount() == 0` was not reckoned with. Followed up that one with an overall code review and fixed all the spots where the code for the `count == 0` scenario was either non-trivial or would potentially produce random/garbage results.

(For example, a zero-length calloc on my system is a hard fail because it's way too often indicative of faulty coding somewhere, so I won't wait for `LEPT_CALLOC` to b0rk with a NULL return. This is more... pedantic than one may expect on Linuxes in general.)

### DIFF
--- a/src/colorquant1.c
+++ b/src/colorquant1.c
@@ -3110,7 +3110,8 @@ PIXCMAP   *cmap;
     makeRGBToIndexTables(level, &rtab, &gtab, &btab);
 
         /* The octarray will give a ptr from the octcube to the colorarray */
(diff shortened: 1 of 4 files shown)


DanBloomberg:
merged; thanks!

PR #709, opened 2023-08-19
fix temporary directory handling on macOS

benoit-pierre:
Similarly to Windows, redirect `/tmp/…` paths accesses to the correct folder on macOS, fixing running the testsuite on the later.

PR #713, opened 2023-08-31
 Use macOS specific directory instead of /tmp 

stweil:
In addition, a trivial compiler warning was fixed by using an ANSI C function prototype.

### DIFF
--- a/src/encoding.c
+++ b/src/encoding.c
@@ -280,7 +280,7 @@ isBase64(char  c)
  * \brief   genReverseTab64()
  */
 static l_int32 *
-genReverseTab64()
+genReverseTab64(void)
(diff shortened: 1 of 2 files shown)


stweil:
These modifications were inspired by pull request #709 which is replaced here.

DanBloomberg:
Thank you.  Really nice to have this resolved.
Good to see the test suite passes on macOS.
All looks good except for comments at lines 2395 and 2345.

stweil:
> All looks good except for comments at lines 2395 and 2345.

Please help me. How should those comments look like?

DanBloomberg:
Not a big deal; can be modified later.

PR #714, opened 2023-09-01
autoconf: Remove obsolete AC_PROG_LIBTOOL

stweil:
It was replaced by LT_INIT (which is already in configure.ac).


DanBloomberg:
Thank you.

PR #715, opened 2023-09-01
Use atomic operations for some index variables

stweil:
This makes the related functions thread safe, but calling them from different processes will still cause problems because all processes use the same filenames which are generated by using the index.


DanBloomberg:
Thanks!

PR #716, opened 2023-09-01
Add GitHub action for build and test on macOS

stweil:


stweil:
See this [example for the new action](https://github.com/stweil/leptonica/actions/runs/6051720690).

DanBloomberg:
Elegant!!

I looked at the example, and saw it did 147 tests.  I've been getting 146 when run locally using alltests_reg.
Comparing the set, I found that numa3_reg.c had been omitted from alltests_reg, for no good reason.  So I'll add it.

stweil:
@DanBloomberg, I wonder whether the new action really should be triggered by every pull request and git push. In addition it can be triggered manually by project maintainers ([url](https://github.com/DanBloomberg/leptonica/actions/workflows/autotools-macos.yml)). Each action run uses two CPU cores for 12 minutes, one for macOS 12 and one for macOS 13.

Maybe it would be sufficient to run only on macOS 13 by default, so macOS 12 would only be tested when that was requested manually. What do you think?

DanBloomberg:
Running by default only with macOS 13 is fine with me.

Considering possible overuse of github resources, what do you think about the fuzzing that is run on every git push?   (I don't know if it is run with PRs)

stweil:
The fuzzing action is only triggered by pull requests, see the [code](https://github.com/DanBloomberg/leptonica/blob/eeefd24998503df9714f921e5398461175ea3896/.github/workflows/cifuzz.yml#L2). It typically runs for about 30 minutes (see [history](https://github.com/DanBloomberg/leptonica/actions/workflows/cifuzz.yml)). That looks like a waste of resources, for example electrical energy, because Google's OSS-Fuzz also runs fuzzers for Leptonica (not only for a few minutes which won't find issues, but really long) and reports build errors by e-mail. @AdamKorcz, what do you think?

PR #718, opened 2023-10-31
Remove redundant assignment (fixes compiler warning)

stweil:
The member variable bpp is deprecated and a synonym for the member variable prec, which already has the same value assigned to it, so remove the redundant assignment to bpp. This fixes a compiler warning:

src/jp2kio.c: In function 'pixConvertToOpjImage':
src/jp2kio.c:726:9: warning: 'bpp' is deprecated: Use prec instead [-Wdeprecated-declarations]
  726 |         cmptparm[i].bpp = 8;
      |         ^~~~~~~~


DanBloomberg:
Thank you for this!

PR #731, opened 2024-02-01
Remove C99 style inline declaration

chris-liddell:
Trivial patch.

### DIFF
--- a/src/adaptmap.c
+++ b/src/adaptmap.c
@@ -2982,10 +2982,12 @@ PIX  *pix1, *pix2, *pixd;
 PIX *
 pixConvertTo8MinMax(PIX  *pixs)
 {
+l_int32  d;
+
     if (!pixs)
         return (PIX *)ERROR_PTR("pixs not defined", __func__, NULL);
 
-    l_int32 d = pixGetDepth(pixs);
+    d = pixGetDepth(pixs);
     if (d == 1) {
         return pixConvert1To8(NULL, pixs, 255, 0);
     } else if (d == 2) {


stweil:
That code was added in commit df0fe14f8.

I am curious: are there still build environments which don't accept such inline declarations (which I personally like because they require less code lines and are easier to review)?

If we want to avoid inline declarations, I suggest to update the commit and format the declaration like most others in the Leptonica code (left aligned, empty line after declaration).

stweil:
C99 is now more that 20 years old. "C99 is substantially completely supported as of GCC 4.5". gcc 4.5.0 was released April 14, 2010.  I think that compilers are no longer a reason to avoid them.

DanBloomberg:
You are probably correct, and perhaps we should not sweat the inline declarations going forward.

Perhaps more important than the date that gcc first supported C99 is the reluctant acceptance of C99 by Microsoft 

This is from Wikipedia:
```
Most C compilers provide support for at least some of the features introduced in C99.

Historically, [Microsoft](https://en.wikipedia.org/wiki/Microsoft) has been slow to implement new C features in their [Visual C++](https://en.wikipedia.org/wiki/Visual_C%2B%2B) tools, instead focusing mainly on supporting developments in the C++ standards.[[13]](https://en.wikipedia.org/wiki/C99#cite_note-13) However, with the introduction of Visual C++ 2013 Microsoft implemented a limited subset of C99, which was expanded in Visual C++ 2015.[[14]](https://en.wikipedia.org/wiki/C99#cite_note-vs2015_new-14)
```

chris-liddell:
Oops, sorry, I did miss the other requirements....

This came up because we (Ghostscript) still get used on some older, obscure platforms (like a Unix emulation environment running on IBM s/390 mainframes) where gcc is not the de facto standard. IBM's C compiler lags behind and, due to the astronomical upgrade costs, their customers lag even further behind. Which is why Ghostscript still avoids C99 and later constructs.

stweil (approved):
Looks good now, thank you!

DanBloomberg:
Thank you for explaining.  Those legacy issues are always annoying.
We'll keep avoiding inline declarations.

PR #732, opened 2024-02-05
Make pix{Read,Write}MemJp2k work via direct memory access

atykhyy:
On platforms lacking `open_memstream()`, `pix{Read,Write}MemJp2k()` worked via a temporary file, but this overhead is unnecessary as openjpeg2000's `opj_stream` interface allows one to read/write directly from a memory buffer by supplying it with appropriate callbacks. This PR makes `pix{Read,Write}MemJp2k()` use this more direct mechanism on all platforms.

In addition, since `pixReadStreamJp2k()` always reads the whole stream into memory when it called `fgetJp2kResolution()`, this PR saves a bit of compute resources by in effect making `pixReadStreamJp2k()` into a wrapper for `pixReadMemJp2k()`.


DanBloomberg:
This is very nice work!
I patched your change and jp2kio_reg passes, so the output files are identical!
Thank you!

PR #740, opened 2024-04-07
cmake: fix library name

Andarwinux:
For CYGWIN/MSYS/MINGW, libraries should be named the same as UNIX.


zdenop (commented):
I am fine with this change

DanBloomberg:
Thank you for the fix, and thank you @zdenop for the review!

PR #753, opened 2024-10-12
Replace sh by bash for reg_wrapper.sh and fix test for working gnuplot

stweil:
Some older versions of /bin/sh don't work.
This was noticed in a test on Solaris 11.4 (SPARC) with sh (AT&T Research) 93u+ 2012-08-01.

Instead of using /bin/bash the script now finds bash with /usr/bin/env, so it will use the latest bash on systems like macOS.

Fix also the test for a working gnuplot.

`gnuplot -e "set terminal png"` does not return an error if the command fails, but prints an error message.

So instead of checking for an error the test now checks for a non-empty output.

This fixes `make check` on hosts with a gnuplot which does not support `set terminal png`.

### DIFF
--- a/prog/reg_wrapper.sh
+++ b/prog/reg_wrapper.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 #
 #  This testing wrapper was written by James Le Cuirot.
 #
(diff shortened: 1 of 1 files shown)


DanBloomberg:
Nice!!
A gnuplot version that doesn't support a png terminal has got to be from the early part of the century  :-)

I'm signing off.  Feel free to push to head when the tests complete.

stweil:
> A gnuplot version that doesn't support a png terminal has got to be from the early part of the century :-)

It's the very latest gnuplot code, but compiled on Solaris without libpng.

PR #758, opened 2024-10-25
Remove broken CMake config files from autotools

### LINKED ISSUE #757: Autotools build installs broken CMake configuration files
The autotools build has been installing broken CMake files as it currently cannot substitute all the variables nor generate a LeptonicaTargets.cmake file. For example, LeptonicaConfig.cmake getting installed with:
```cmake
if (@OPENJPEG_SUPPORT@)
    find_dependency(OpenJPEG CONFIG)
endif()
if (@LIBWEBP_SUPPORT@)
    find_dependency(WebP @MINIMUM_WEBPMUX_VERSION@ CONFIG)
endif()

include(${CMAKE_CURRENT_LIST_DIR}/LeptonicaTargets.cmake)

...

SET(Leptonica_VERSION           @leptonica_VERSION@)
SET(Leptonica_VERSION_MAJOR     @leptonica_VERSION_MAJOR@)
SET(Leptonica_VERSION_MINOR     @leptonica_VERSION_MINOR@)
SET(Leptonica_VERSION_PATCH     @leptonica_VERSION_PATCH@)
```

May just remove these files from Autotools build if they aren't supported. 

https://github.com/DanBloomberg/leptonica/blob/96a3d7451e7d717d8a0c88436f5ff7ea7129412e/configure.ac#L270-L271

https://github.com/DanBloomberg/leptonica/blob/96a3d7451e7d717d8a0c88436f5ff7ea7129412e/Makefile.am#L10-L12

cho-m:
To fix #757 by removing the broken files. It isn't simple to generate CMake config target file. Could store a generated copy from CMake, but would need to make sure it doesn't go out-of-sync and regenerate it as necessary.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -5,12 +5,6 @@ EXTRA_DIST = config README.html leptonica-license.txt moller52.jpg version-notes
 pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = lept.pc
 
-# Cmake configs:
-lept_cmakeconfigdir = $(libdir)/cmake/
-lept_cmakeconfig_DATA = \
-	cmake/templates/LeptonicaConfig.cmake \
-	cmake/templates/LeptonicaConfig-version.cmake
-
 SUBDIRS = src prog
 
 # The fuzzing tests are run by OSS-Fuzz https://oss-fuzz.com/,
--- a/configure.ac
+++ b/configure.ac
@@ -267,8 +267,6 @@ AC_SUBST([VERSION_PATCH], [$(echo "$AX_POINT_VERSION" | $SED 's/^\([[0-9]][[0-9]
 INCLUDE_PATH=`eval echo $includedir`
 AC_SUBST([INCLUDE_DIR], ["${INCLUDE_PATH}/leptonica"])
 AC_SUBST([leptonica_OUTPUT_NAME], ["leptonica"])
-AC_CONFIG_FILES(cmake/templates/LeptonicaConfig.cmake)
-AC_CONFIG_FILES(cmake/templates/LeptonicaConfig-version.cmake)
 
 # Create symlink
 AC_PROG_LN_S


zdenop:
if something is broken - it should be fixed not removed. Otherwise, you finish with an empty disk.

cho-m:
> if something is broken - it should be fixed not removed. Otherwise, you finish with an empty disk.

It never worked since original commit https://github.com/DanBloomberg/leptonica/commit/bab57defd75ebd02b0fdb832dad22222d52273b9 had same include:
https://github.com/DanBloomberg/leptonica/blob/2daaaad77992882192faa6b48bf0a8f5df99aa75/cmake/templates/LeptonicaConfig.cmake.in#L31

Personally would consider broken functionality negative value while removed feature zero value. A broken feature is only a source of user issues and complaints.

---

The biggest issue is having to deal with sync-ing of CMake and Autotools. Without proper CI to guarantee functionality of generated CMake config files are identical, it is highly likely they will break again in the future.

Ignoring that, a simplified target file could be created for current single target. However, it will be very fragile and can break on slightest modification of CMake target (adding a new dependency, adding a new target, renaming something, etc)

https://github.com/DanBloomberg/leptonica/blob/2daaaad77992882192faa6b48bf0a8f5df99aa75/src/CMakeLists.txt#L78

zdenop:
So you suggest that it is not worthy to try help those who would like to use cmake in their project, but their distribution is using automake?

cho-m:
> So you suggest that it is not worthy to try help those who would like to use cmake in their project, but their distribution is using automake?

I am just pointing out maintainer burden/commitment on keeping it functional. There is worth is having a stable, functional feature of CMake files. Would strongly recommend test cases if you want to go this route.

Ultimately this is just a recommendation and you are free to supersede it with your own PR. I am fine with closing this and handling it downstream as I was working out the bugs seen in both Autotools and CMake options in Homebrew.

PR #760, opened 2024-10-25
Add CTest support and option to skip test progs

### LINKED ISSUE #759: CMake build with `BUILD_PROG=ON` fails on Linux due to libm
Was experimenting with CMake build at Homebrew https://github.com/Homebrew/homebrew-core/pull/194875 (mainly due to #757) and noticed the Linux build kept failing on libm.

Probably similar to BSD (https://github.com/DanBloomberg/leptonica/commit/d5c36c877f0abe218fa940beecb9ac43dac9041d).

One note is that failure is happening in a file not specified in list, i.e. `blend4_reg`
```
  /usr/bin/ld: CMakeFiles/blend4_reg.dir/blend4_reg.c.o: undefined reference to symbol 'sqrt@@GLIBC_2.2.5'
  /usr/bin/ld: /lib/x86_64-linux-gnu/libm.so.6: error adding symbols: DSO missing from command line
```
https://github.com/DanBloomberg/leptonica/blob/96a3d7451e7d717d8a0c88436f5ff7ea7129412e/prog/CMakeLists.txt#L2

Can also see Fedora's workaround to always use `-lm` - https://src.fedoraproject.org/rpms/leptonica/blob/rawhide/f/leptonica_cmake.patch#_41

---

Many of these look like test programs, so perhaps putting them behind a CMake option so that users can build only the `INSTALL_PROGS` would be useful.

cho-m:
Partially fix #759 when building for distribution that doesn't care about test programs. Similar idea to how CTest `-DBUILD_TESTING=OFF` or GNOME's common meson `-Dtests=disabled` usually avoids compiling unnecessary files.

Build can still fail on Linux/glibc when `-DBUILD_TESTING=ON` as some progs needs `-lm` but lower priority for packaging which is my main concern.

Used `BUILD_TESTING` option as that is default for CTest[^1] so can be compatible if future switch to `include(CTest)`.

Note that CI does not test this as `BUILD_PROG=OFF` for Linux:
https://github.com/DanBloomberg/leptonica/blob/96a3d7451e7d717d8a0c88436f5ff7ea7129412e/.github/workflows/cmake.yml#L46

[^1]: https://cmake.org/cmake/help/latest/module/CTest.html

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -48,4 +48,8 @@ jobs:
     - name: Build
       # Build your program with the given configuration
       run: sudo cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --target install
+
+    - name: Test
+      if: runner.os == 'Linux'
+      run: ctest --test-dir ${{github.workspace}}/build
   
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,7 @@ project(
   leptonica
   LANGUAGES C
   VERSION 1.85.1)
+include(CTest)
 set(CMAKE_C_STANDARD 11)
 set(CMAKE_C_STANDARD_REQUIRED ON)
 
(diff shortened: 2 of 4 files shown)


cho-m:
Not entirely sure on test cases. Seems to run.

However, I think the full test suite will fail due without `gnuplot`

zdenop:
The primary building tools are autotools. I am not sure if it make sense to run the same with another build system (cmake). This is the reason why did not build progs with cmake (not to waste resources and time).

PR #761, opened 2024-11-02
cmake.yml: test building programs

cho-m:
Try building these to see if #759 can be reproduced on Linux runner.


DanBloomberg:
two apparently different failures when attempting to build:
ubuntu: couldn't find sqrt (from libm) to link
macos:  failed earlier; couldn't find system include files like stdio.h

DanBloomberg:
Thanks, Zdenko!

PR #771, opened 2025-05-17
fix issue 765 and improve cmake configuration

zdenop:


DanBloomberg:
Thank you, Zdenko -- that was a lot of work!

One question: for some reason it is trying to build leptonica fuzzers, and failing.  I can't think of why it needs to build those fuzzers.

zdenop:
base on original suggestion https://github.com/DanBloomberg/leptonica/commit/0a278248966d979e38817068c607c358b16a0516#r60209158 and tesseract I modified CI fuzzer file.  I hope this will helps.

PR #773, opened 2025-07-06
Fix some issues reported by Coverity Scan

stweil:


DanBloomberg:
Not sure what C.S. was complaining about with the values of ascii85flag.  And it's used several other times in the file, in apparently the same way, so why didn't C.S. complain in those places?

DanBloomberg:
Stefan, did you want to check this in, or make any modifications?
Whatever you decide is fine with me.

stweil:
Hi Dan, sorry that it took me so long to answer. Maybe Coverity Scan did not complain for the other similar locations because there the validation check for 0 or 1 was missing. Commit 25236783a312b569407cf18a611969f955ee19e6 now adds such checks and uses the same pattern in the if statements.

DanBloomberg:
Thanks for fixing these!

PR #781, opened 2025-10-28
cmake: add option to disable cmake file installation

FrozenGalaxy:


stweil:
@FrozenGalaxy, why do you think this option is necessary? Is there a specific reason why you need it?

PR #791, opened 2026-05-06
CMake: honor ENABLE_* for optional deps and HAVE_* in config

uilianries:
Greetings! 

As Leptonica is not only available as a public project, but also packaged via many package managers, including in [Conan Center](https://conan.io/center/recipes/leptonica?version=1.87.0), sometimes we need to patch some behavior, which we prefer sending to the upstream as a fix/suggetion.

Currently, the project uses CMake to find and configure dependencies in two ways: via the user's options like `ENABLE_ZLIB` (which is great), and the automatic configuration from `cmake/Configure.cmake`, which is based on CMake results, not in options.

The problem is CMake remembers those `_FOUND` paths, and sometimes old `HAVE_*` state in `CMakeCache.txt`. If someone configures with zlib, then re-runs with `-DENABLE_ZLIB=OFF`, the find block may not run again, but `ZLIB_FOUND` can still be true in the cache, not following what was asked.

This PR proposes a more management: For `SW_BUILD=OFF`, link each optional dependency only when the matching **`ENABLE_*`** is on and the existing `*_LIBRARIES` checks succeed (same variable-based linking as before), so we prioritize the user option and avoid CMake cache problems. 

Regards!


src/CMakeLists.txt
```diff
@@ -32,36 +32,35 @@ if(BUILD_SHARED_LIBS)
   target_compile_definitions(leptonica PRIVATE -DLIBLEPT_EXPORTS)
 endif()
 
-if(GIF_LIBRARIES)
+if(NOT SW_BUILD AND ENABLE_GIF AND GIF_LIBRARIES)
```
zdenop: Please put it one condition as in https://github.com/uilianries/leptonica/blob/a6b578471884e4215860fe43156854da9393bff8/cmake/Configure.cmake#L89
```
if(NOT SW_BUILD)
...
endif()
```

uilianries: Done via commit [59e91cc](https://github.com/DanBloomberg/leptonica/pull/791/commits/59e91ccf88b26c5f157798a70a80c98648f4edfa)

zdenop:
@egorpugin : sw (linux, mac, win) is [failing](https://github.com/DanBloomberg/leptonica/actions/runs/25481138084?pr=791). Can you have a look at this?

zdenop:
Thanks

uilianries:
Thank you all for your support and review! :rocket:

PR #792, opened 2026-06-03
Fix use-after-free in morphSequenceVerify()

cjhxmx:
Remove reference to freed 'op' pointer in error message at line 845. The pointer is freed at line 841 in the last loop iteration, then used in the post-loop error check. Replace with a fixed error string that does not reference 'op'.


DanBloomberg:
Thanks!

DanBloomberg:
Thanks!

PR #793, opened 2026-06-03
Fix/fpix1 dpix error and format

vidiecan:
## Summary 

Two low-risk bugs in the FPix/DPix size-consistency check in `src/fpix1.c`.

**1. `dpixReadStream` typo.** The DPix reader returned `ERROR_PTR("invalid fpix data size", …)` — copy-pasted from the fpix path → now `"invalid dpix data size"`.

**2. `%llu` / `l_int32` mismatch (both readers).** The size-consistency `L_ERROR` passes `nbytes`/`w`/`h` (`l_int32`) to `%llu` specifiers, which expect `unsigned long long` — a variadic type mismatch, i.e. undefined behavior in C (ISO C §7.21.6.1p9). `expected` was already `(l_uint64)`; the other three are now cast to match. These calls route through the unannotated `lept_stderr`, so no `-Wformat` / MSVC C4477 warning fires today — it's latent UB, not a build break.

## Behaviour

No behaviour change on valid input (the size-consistency `L_ERROR` runs only when `expected != nbytes`); only the error message printed for malformed input is corrected — its `dpix` wording and its printed numbers.


DanBloomberg:
Thanks for submitting this PR.
The (l_uint64) casts are correct, and there is a typo.

The other issues (upstream merges) are with respect to versions that are far out of date (about 7 years old) and, FWICT, have already been fixed.

I'll submit these changes and we will not merge this PR.

DanBloomberg:
Thanks for doing this.
Those merges are with 7 year old versions, and the issues were addressed long ago.
The casts and typo have been fixed by hand.

PR #794, opened 2026-06-11
ci: add permissions and pin third-party action SHAs

Alb3e3:
> **AI assistance disclosure:** This pull request was prepared with AI assistance. The changes are mechanical CI/CD hardening; I reviewed and validated each workflow change myself, and I take responsibility for the contribution.

---

## Summary

This narrows GitHub Actions workflow privileges and pins only non-provider third-party actions to immutable revisions.

**Changes:**
- Add explicit `permissions: contents: read` to build/fuzzing workflows that only need repository read access.
- Pin `ilammy/setup-nasm` and `egorpugin/sw-action` to commit SHAs because they are third-party actions.
- Leave GitHub-owned actions, Google OSS-Fuzz actions, and the Fedora container tag on their existing tags per maintainer preference for trusted provider updates.

I verified the workflow YAML still parses and `git diff --check` passes. I also re-ran `zizmor`; the remaining high-confidence findings are the intentionally unpinned trusted provider refs and `fedora:latest` tag discussed in review.

### DIFF
--- a/.github/workflows/autotools-macos.yml
+++ b/.github/workflows/autotools-macos.yml
@@ -23,6 +23,9 @@ on:
   # Trigger workflow in GitHub web frontend or from API.
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
 
   brew:
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -10,6 +10,9 @@ on:
     - '**.h'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
--- a/.github/workflows/cmake-win64.yml
+++ b/.github/workflows/cmake-win64.yml
@@ -12,12 +12,15 @@ on:
 env:
   ILOC: d:/a/local
 
+permissions:
+  contents: read
+
 jobs:
   build:
     name: cmake-win64
     runs-on: windows-latest
     steps:
-      - uses: ilammy/setup-nasm@v1
+      - uses: ilammy/setup-nasm@72793074d3c8cdda771dba85f6deafe00623038b  # v1
       - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
         uses: actions/checkout@v7
         with:
--- a/.github/workflows/sw.yml
+++ b/.github/workflows/sw.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@v7
-    - uses: egorpugin/sw-action@0261dc5095b3c7688d7ad1a429ea53d9b239ad6f  # master
+    - uses: egorpugin/sw-action@master
     - run: ./sw build -static -shared -config d,r
     
   linux:
@@ -35,14 +35,14 @@ jobs:
       run: |
         sudo dnf -y update
         sudo dnf -y install cmake which g++ clang clang-tools-extra lld
-    - uses: egorpugin/sw-action@0261dc5095b3c7688d7ad1a429ea53d9b239ad6f  # master
+    - uses: egorpugin/sw-action@master
     - run: ./sw build -static -shared -config d,r
    
   macos:
     runs-on: macos-latest
     steps:
     - uses: actions/checkout@v7
-    - uses: egorpugin/sw-action@0261dc5095b3c7688d7ad1a429ea53d9b239ad6f  # master
+    - uses: egorpugin/sw-action@master
     - name: install
       run: |
         brew update
--- a/.github/workflows/sw.yml
+++ b/.github/workflows/sw.yml
@@ -15,12 +15,15 @@ on:
 #      - '.github/*'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   windows:
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@v7
-    - uses: egorpugin/sw-action@master
+    - uses: egorpugin/sw-action@0261dc5095b3c7688d7ad1a429ea53d9b239ad6f  # master
     - run: ./sw build -static -shared -config d,r
     
   linux:
@@ -32,14 +35,14 @@ jobs:
       run: |
         sudo dnf -y update
         sudo dnf -y install cmake which g++ clang clang-tools-extra lld
-    - uses: egorpugin/sw-action@master
+    - uses: egorpugin/sw-action@0261dc5095b3c7688d7ad1a429ea53d9b239ad6f  # master
     - run: ./sw build -static -shared -config d,r
    
   macos:
     runs-on: macos-latest
     steps:
     - uses: actions/checkout@v7
-    - uses: egorpugin/sw-action@master
+    - uses: egorpugin/sw-action@0261dc5095b3c7688d7ad1a429ea53d9b239ad6f  # master
     - name: install
       run: |
         brew update
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -13,6 +13,9 @@ env:
   # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
   BUILD_TYPE: Release
 
+permissions:
+  contents: read
+
 jobs:
   build:
     # The CMake configure and build commands are platform agnostic and should work equally well on Windows or Mac.
@@ -53,4 +56,3 @@ jobs:
     - name: Build
       # Build your program with the given configuration
       run: sudo cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --target install
-


.github/workflows/sw.yml
```diff
@@ -15,31 +15,33 @@ on:
...
+    - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
+    - uses: egorpugin/sw-action@0261dc5095b3c7688d7ad1a429ea53d9b239ad6f  # master
     - run: ./sw build -static -shared -config d,r
     
   linux:
     runs-on: ubuntu-22.04
     container: fedora:latest
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
```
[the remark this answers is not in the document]
Alb3e3: Addressed in 6cfd730: sw.yml now uses fedora:42@sha256:99e203b80b1c3d8f7e161ec10a68fd02b081ef83a3963553e513c82846b97814 instead of fedora:latest. I also re-ran YAML parsing, git diff --check, and zizmor with medium severity/confidence locally after the follow-up change.

stweil:
@Alb3e3, did you use AI for this pull request?

.github/workflows/autotools-macos.yml
```diff
@@ -36,7 +39,7 @@ jobs:
           - { os: macos-latest, cc: clang }
 
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
```
stweil: Isn't this an action which is provided by the owners of GitHub (Microsoft)? Do they propose to harden their own actions?

.github/workflows/cifuzz.yml
```diff
@@ -10,25 +10,28 @@ on:
...
+  contents: read
+
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
     steps:
     - name: Build Fuzzers
       id: build
-      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@439b356d09b53fe1161e9fe22ac42536926983db  # master
```
stweil: This is an action which is provided by Google. Do they propose to harden their own actions? Or do they suggest to use their latest code with the latest fixes and improvements?

Alb3e3:
Hi @stweil — yes, I used AI to help draft and double-check this, but the changes and the reasoning behind them are mine. It's a fairly mechanical hardening PR: adding top-level `permissions` and pinning the third-party actions to commit SHAs across the six workflows. I went through each workflow myself to confirm the pinned SHAs match the tags they replace and that nothing breaks the existing jobs. Happy to adjust anything — e.g. narrowing `read-all` to explicit `contents: read` per workflow if you'd prefer that.

stweil:
@Alb3e3, I think that contributions which used AI should clearly state this. Leptonica and Tesseract still have no AI related coding guidelines, but have a look at the [guidelines for Koha](https://wiki.koha-community.org/wiki/Coding_Guidelines#AI_and_LLM-assisted_contributions) which show how AI contributions should be honoured.

My human intelligence thinks that this pull request has several issues (see also my remarks above), and when I ask Claude.AI, it comes to the same result. I won't go into details here because you can think yourself or ask an AI of your choice.

Alb3e3:
Thanks @stweil — that's fair. I've added an explicit AI-assistance disclosure to the PR description. To be clear: AI helped me draft and cross-check the changes, but they're mechanical workflow hardening (top-level `permissions` + pinning actions to commit SHAs), I reviewed each one against the actual workflows, and I stand behind the contribution. Happy to adjust to whatever AI-disclosure convention you and the Tesseract project settle on.

.github/workflows/autotools-macos.yml
Alb3e3: Good point. GitHub-owned actions are a different trust/maintenance tradeoff from arbitrary third-party actions. The reason I pinned this one was reproducibility against mutable tag movement, not because Microsoft specifically asks projects to harden `actions/*` this way.

If Leptonica prefers to treat GitHub-owned actions as trusted and keep them on version tags for easier updates, I can re-scope the PR accordingly rather than argue for SHA pins there.

.github/workflows/cifuzz.yml
Alb3e3: Same tradeoff here: the pin is meant to avoid unreviewed tag movement and make the exact reviewed code reproducible. It is not based on Google requiring projects to pin Google's own action.

For Google-maintained actions there is a reasonable counterargument that a version/tag can receive upstream fixes more conveniently. If you prefer that policy for `google/oss-fuzz` actions, I can narrow the PR to the permission changes and any pins you are comfortable maintaining.

.github/workflows/autotools-macos.yml
stweil: The whole infrastructure for GitHub actions including the runners is provided by Microsoft, so without trust Leptonica would have to stop using it. Their GitHub actions are part of this infrastructure and can be trusted, too. SHA pins would require more updates, so I would not use them here.

.github/workflows/cifuzz.yml
stweil: I think that Google can be trusted, too, and google/oss-fuzz is not some random project, but a well maintained software repository. SHA pins for actions from trusted parties would require more updates, so I would not use them here.

.github/workflows/sw.yml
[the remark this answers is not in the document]
stweil: My comments about trusted parties also apply to fedora. Fixing the container to a certain SHA256 would prevent getting updates with bug fixes, including security fixes.

stweil:
@Alb3e3, did you read the review from Copilot? I think that using SHA256 should be restricted to actions from untrusted third parties (update your commit message to make this clear). And the permissions should be reviewed, may handled in a separate pull request.

Alb3e3:
Thanks, I reworked the branch to match that policy direction.

Updated in `5e1cd1b`:
- reverted SHA pins for GitHub-owned actions (`actions/checkout`, `actions/upload-artifact`, `github/codeql-action`)
- reverted SHA pins for the Google OSS-Fuzz CI actions
- reverted the Fedora container digest back to `fedora:latest`
- kept SHA pins only for non-provider third-party actions: `ilammy/setup-nasm` and `egorpugin/sw-action`
- kept explicit `permissions: contents: read` on the build/fuzzing workflows for review
- updated the commit message, PR title, and PR body so the scope is clear

Local validation: workflow YAML parses and `git diff --check` passes. I also re-ran `zizmor`; the remaining high-confidence findings are the intentionally unpinned trusted provider refs and `fedora:latest` tag discussed above.

Alb3e3:
Rebased onto current master to clear the merge conflict (the `sw.yml` / checkout v6→v7 changes that landed in the meantime). No content changes from the last revision: `permissions: contents: read` on the build/fuzz workflows, SHA pins kept only for the third-party actions (`ilammy/setup-nasm`, `egorpugin/sw-action`), GitHub-owned actions left on version tags.

.github/workflows/sw.yml
```diff
@@ -15,12 +15,15 @@ on:
...
+permissions:
+  contents: read
+
 jobs:
   windows:
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@v7
-    - uses: egorpugin/sw-action@master
+    - uses: egorpugin/sw-action@0261dc5095b3c7688d7ad1a429ea53d9b239ad6f  # master
```
stweil:
```suggestion
    - uses: egorpugin/sw-action@master
```

.github/workflows/sw.yml
```diff
@@ -32,14 +35,14 @@ jobs:
       run: |
         sudo dnf -y update
         sudo dnf -y install cmake which g++ clang clang-tools-extra lld
-    - uses: egorpugin/sw-action@master
+    - uses: egorpugin/sw-action@0261dc5095b3c7688d7ad1a429ea53d9b239ad6f  # master
```
stweil:
```suggestion
    - uses: egorpugin/sw-action@master
```

.github/workflows/sw.yml
```diff
@@ -32,14 +35,14 @@ jobs:
...
     - run: ./sw build -static -shared -config d,r
    
   macos:
     runs-on: macos-latest
     steps:
     - uses: actions/checkout@v7
-    - uses: egorpugin/sw-action@master
+    - uses: egorpugin/sw-action@0261dc5095b3c7688d7ad1a429ea53d9b239ad6f  # master
```
stweil:
```suggestion
    - uses: egorpugin/sw-action@master
```

stweil (changes requested):

.github/workflows/cifuzz.yml
```diff
@@ -10,6 +10,9 @@ on:
     - '**.h'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
```
stweil, line 15: What about the upload of artifacts? Will it still work?

.github/workflows/cmake-win64.yml
stweil, line 171: What about the upload of artifacts? Will it still work with the new permissions?

.github/workflows/cifuzz.yml
Alb3e3: I checked this against `actions/upload-artifact@v7`. The upload path does not use the repository-scoped `GITHUB_TOKEN` permission set by this workflow-level `permissions` block. The action is backed by `@actions/artifact`, which reads `ACTIONS_RUNTIME_TOKEN` and `ACTIONS_RESULTS_URL` and calls the Actions Results service (`CreateArtifact` / `FinalizeArtifact`) with that runtime token.

So `permissions: contents: read` should not block `actions/upload-artifact`; it only narrows the repository token available to ordinary workflow steps. I re-checked workflow YAML parsing and `git diff --check`. I did not add `actions: write` because that would broaden the repo token without being required for artifact upload.

.github/workflows/cmake-win64.yml
Alb3e3: I checked this against `actions/upload-artifact@v7`. The upload path does not use the repository-scoped `GITHUB_TOKEN` permission set by this workflow-level `permissions` block. The action is backed by `@actions/artifact`, which reads `ACTIONS_RUNTIME_TOKEN` and `ACTIONS_RESULTS_URL` and calls the Actions Results service (`CreateArtifact` / `FinalizeArtifact`) with that runtime token.

So `permissions: contents: read` should not block `actions/upload-artifact`; it only narrows the repository token available to ordinary workflow steps. I re-checked workflow YAML parsing and `git diff --check`. I did not add `actions: write` because that would broaden the repo token without being required for artifact upload.

stweil:
Thanks for your contribution and for your patience!

PR #796, opened 2026-06-22
Add initial Dependabot configuration

stweil:
This configuration will automatically create pull requests for updates of GitHub actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  # Update GitHub actions.
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


DanBloomberg:
Thanks Stefan.  This looks like an elegant way to handle the actions.

stweil:
Dan, please make also the necessary advanced security settings for the repository in GitHub. "Dependabot version updates" is needed here. These settings can only be made by project owners.

Bildschirmfoto 2026-06-23 um 05 52 52

stweil:
Oh, it looks like it's already working: PR #797 is a result of the new configuration.

DanBloomberg:
#796 set up the dependabot version updates, exactly as in the documentation.

Here's what I see on advanced security.  These dependabot updates don't appear to be enabled, unlike what you showed.

github-adv-security-page

stweil:
So the last four settings (starting from "Dependabot malware alerts") should be enabled.

DanBloomberg:
I enabled the first 3.

When I 'configure' for the version updates, it shows me the dependabot.yml file, but I don't want to edit it.  And the security page continues to say "Configure" for that item.  It's already configured.

I had a long dialog with Google search on this and do not understand how to fix this.  I will send you that dialog separately.

PR #802, opened 2026-06-30
ci: Remove unused ilammy/setup-nasm

stweil:
A positive side effect of this removal is that it not only simplifies the CI rules, but also fixes this CI warning:

    Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: ilammy/setup-nasm@v1.

### DIFF
--- a/.github/workflows/cmake-win64.yml
+++ b/.github/workflows/cmake-win64.yml
@@ -20,7 +20,6 @@ jobs:
     name: cmake-win64
     runs-on: windows-latest
     steps:
-      - uses: ilammy/setup-nasm@72793074d3c8cdda771dba85f6deafe00623038b  # v1
       - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
         uses: actions/checkout@v7
         with:


zdenop:
AFAIR nasm is needed for building libjpeg

stweil:
Building libjpeg works without nasm (see [build log](https://github.com/stweil/leptonica/actions/runs/28443650468/job/84287417769)). It only reports a warning:
```
-- Looking for a ASM_NASM compiler - NOTFOUND
CMake Warning at simd/CMakeLists.txt:5 (message):
  SIMD extensions disabled: could not find NASM compiler.  Performance will
  suffer.
Call Stack (most recent call first):
  simd/CMakeLists.txt:57 (simd_fail)
```

Is this relevant? We don't want to test builds of libjpeg-turbo, but it is part of the created artifact archive file. I think it would be acceptable to have a libjpeg-turbo without SIMD optimizations there.

DanBloomberg:
seems ok to me not to require using the SIMD compiler. with libjpeg

